PR #53, opened 2014-04-29
Make the project function more like a typical ruby gem

coop:
At [EverydayHero](https://github.com/everydayhero) we've been long time users of Puppet but we haven't participated much in the community. It's only recently that we've started pulling code from the forge and really exploring the code and tools being written by others - in the past we've hand rolled a lot of implementations because we originally launched on 0.24.x when there wasn't a whole lot around.

In a recent project I've had the opportunity to start fresh and am really liking the ideas this project in particular (`puppet-librarian` is another). I read the README and was surprised that `puppetlabs_spec_helper` seems to have dropped rubygems support (deleting the gemspec) which makes it impossible to use with bundler.

I had a decent amount of experience with ruby and I've gone through and refactored the gem to conform to the typical rubygems format - there are a few breaking API changes which means if this was included we'd be looking the bump the version to 1.0.0.

Some of the biggest changes include:

* add the gemspec back - it was removed in a previous commit because the logic was moved into the `Rakefile`. Without a gemspec the code isn't considered a gem which makes it unusable with bundler etc.
* get the specs to run - I couldn't find a way to run the tests. I've modified the `Rakefile` so that `bundle exec rake` runs all the tests.
* got rid of the manual `requires` in most files - the ruby convention is to have a file under `lib/project_name` called `project_name` that lists all the files you'd like to require.
* removed `spec/watchr.rb` because tools like [`guard`](https://github.com/guard/guard) have superseded it. If this is required functionality I'm sure we can figure something out.

What I'd like to see from here:

* Target newer versions of puppet only - `lib/puppetlabs_spec_helper/puppet_spec_helper.rb` is a good example of the hacks that are going on.
* Increase the test coverage - the tests pass but I have no idea if anything else is broken.
* Add the project to travis-ci - this will encourage all new development to include tests and we'll know straight off the bat how the new code impacts existing code.
* Remove dependency on `mocha` - rspec has a perfectly file expectation library which should be used instead (and is already included in the project...)
* Better documentation on how this works with `puppet-rspec` and `puppet-librarian`.

At the moment I'm looking for feedback because I understand that the changes I've made are probably more drastic than could be merged straight up.


hunner:
I think I have incorporated all of the pertinent changes given here. Thanks!

PR #87, opened 2015-01-27
fix load issue with puppet filesystem and windows symlinks

logicminds:
* previously the code was trying to load puppet/filesystem which does not exist
    or cannot be loaded.  This causes windows symlink to raise an error when it shouldn't.
    Because the LoadError is caught the functionality of detecting if the puppet symlink code
    is available fails and the patch to make symlinks work on windows never gets executed.
    By just loading puppet we can still access the puppet/filesystem code and be able to use
    windows symlinks.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -93,7 +93,7 @@ task :spec_prep do
   is_windows = !!File::ALT_SEPARATOR
   puppet_symlink_available = false
   begin
-    require 'puppet/file_system'
+    require 'puppet'
     puppet_symlink_available = Puppet::FileSystem.respond_to?(:symlink)
   rescue
   end


logicminds:
Furthermore, the current gem spec does not include puppet as a runtime dependency.  I am not sure if this is needed because it is expected you to have puppet already installed.  However, for testing procedures  to work correctly we may need to add puppet as a development dependency.

hunner:
What version of puppet are you using? `puppet/file_system` does appear to be available on all of my unit tests.

Puppet is not declared as a dependency because that would force the gem to be used rather than allowing native system packages.

hunner:
Looking through git, it appears to be introduced in puppet 3.2.4

cyberious:
The whole reason this was in a begin is because it is available only post 3.2.4 and resolves issues on windows. If you want to add puppet I am fine but you can not remove that include.

cyberious:
Also there is no reason to load the whole puppet library if we know this function will only work if the 'puppet/file_system'

cyberious:
Found the same failure against 3.7.4. Merging PR, not sure what file_system is now depending on in core that is causing it to not load.

PR #89, opened 2015-02-04
Enable strict_variables by default if puppet version > 4.0

mcanevet:
This may break a lot of tests but could bring out a lot a bugs.
Few people actually uses STRICT_VARIABLES in its matrix travis (which is bad), so let's force it...

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -101,8 +101,8 @@ variables for the spec run. These are:
 * ``FUTURE_PARSER`` - set to "yes" to enable the [future parser](http://docs.puppetlabs.com/puppet/latest/reference/experiments_future.html),
   the equivalent of setting [parser=future](http://docs.puppetlabs.com/references/latest/configuration.html#parser)
   in puppet.conf.
-* ``STRICT_VARIABLES`` - set to "yes" to enable strict variable checking,
-  the equivalent of setting [strict_variables](http://docs.puppetlabs.com/references/latest/configuration.html#strictvariables)=true
+* ``STRICT_VARIABLES`` - set to "no" to disable strict variable checking when using Puppet >= 4.0,
+  the equivalent of not setting [strict_variables](http://docs.puppetlabs.com/references/latest/configuration.html#strictvariables)
   in puppet.conf.
 * ``ORDERING`` - set to the desired ordering method ("title-hash", "manifest", or "random")
   to set the order of unrelated resources when applying a catalog. Leave unset for the default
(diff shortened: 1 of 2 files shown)


nanliu:
Since this is already available as a setting, wouldn't it be better to configure strict checking instead of imposing this globally?

mcanevet:
@nanliu fixed

hunner:
As per PUP-481, this will not be the default until puppet 5, so I'm not going to enforce this yet.

raphink:
Even if it's not activated by default in Puppet itself, it's probably safer to activate it in module tests to be on the safe side. We do it in all of our modules, without side effects (besides finding bugs in our code when we activated it).

daenney:
I think this is a bad idea. As much as it could help us forward be default force-enabling a feature that is not enabled for that same version of Puppet will cause a lot of churn for people and work. Considering strict variables isn't going to default to on before Puppet 5 this is premature.

mcanevet:
The thing is that you may not test the right thing if you don't initialize properly in your unit tests all the facts you are using in your manifests. Which may hide bugs. This will greatly help reducing such errors by failing on uninitialized fact variable.

daenney:
But it's not just uninitialised fact variables. It's uninitialised anything which when you access it in Puppet now will simply give you an `undef` instead of blow up.

DavidS:
@daenney does it help pointing out that the commit should be titled "Enable strict_variables by default" instead of "Always enable ..." ?

I am partial to more strict testing, but I also understand people's needs to move forward at their own pace. Changing the default (as happens here) does support both parts.

daenney:
The title change doesn't help. I don't think we should force defaults on people that are contrary to what is currently required of a module. I love strict variables, I've even made puppetlabs/apt strict variables compliant, as far back as Puppet 3.2, but that was such a freaking HUGE headache if you have to support a multitude of Puppet versions that I'd honestly not wish that on anyone, at least not before Puppet 4.

DavidS:
Would using puppet 4 as cut-off for changing the default work better for
you then? Assuming people need to change their modules then anyways? Adding
a note to that effect, pointing out the possibility to disable?

daenney:
Yes, I've actually expanded on my comment a bit to include that as a suggestion but Github isn't too great with getting those updates out per mail unfortunately.

I think Puppet 4 would be a far more sensible cutoff point.

DavidS:
:+1:

raphink:
Note also that in Puppet 4, strict variables will apply not only to manifests, but to templates and functions as well. I've had quite a few surprises, especially because the new `throw` behavior in Puppet makes it particularly unuseful to debug…

mcanevet:
@DavidS @daenney is it good for you now that I enable strict_variable by default for Puppet4 only?

daenney:
Yap, wfm.

DavidS:
I cherry-picked and pushed this as [6175e5a](https://github.com/puppetlabs/puppetlabs_spec_helper/commit/6175e5a5dc316e1788e2295145b09c508a7457d4).

Thanks for the contribution!

mcanevet:
thanks @DavidS

PR #91, opened 2015-02-11
Enable future parser testing

cmurphy:
rspec-puppet#183 was merged and released in 2.0, so we can now take
advantage of the future parser.


hunner:
This will most definitely cause puppetlabs_spec_helper to require rspec-puppet 2.x, which will be a backwards incompatible change. I know we are 0.x, but still...

hunner:
Err, as pointed out irl FUTURE_PARSER=yes hasn't worked ever since `import` was deprecated.

PR #92, opened 2015-02-24
Remove hard dependency on json

hunner:
The metadata-json-lint gem requires json which is a native extention.
This means it requires a full dev chain to install
puppetlabs_spec_helper with this dependency.

This commit removes the dependency, but instead prints a warning when
the gem is not found but does not error.


nibalizer:
:+1:

PR #96, opened 2015-03-16
Only use --depth 1 if ref is not used

treydock:
The use of "--depth 1" when cloning a git repo makes reseting to a tag not work.  The other option would be to execute `git fetch --tags origin` to continue supporting git cloning of specific tags.

Issue - https://tickets.puppetlabs.com/browse/MODULES-1861


bodepd:
this probably needs a quick merge and a release, I'm guessing this leaves a large number of users in a broken state. I can verify if merging this resolves my build issues.

sbadia:
your option is ok, thanks for the quick fix!

bodepd:
I validated this against my .fixture.yaml file and it seems to have corrected the issue.

bodepd:
thanks for the quick response and update here. I have verified 10.1 in travis.

PR #99, opened 2015-03-27
(MODULES-1206) Added code coverage through SimpleCov for Ruby >= 1.9

leoarnold:
The `rake` task `:coverage` used the `rcov` gem which
> [...] __does not__ work on Ruby 1.9.x.  For coverage on Ruby 1.9 look at [SimpleCov](https://github.com/colszowka/simplecov) ([Quote](https://github.com/relevance/rcov/blob/master/README.markdown))

I fixed this by
* Renaming the `rake` task `:coverage` to `coverage:rcov`
* Adding a `rake` task `coverage:simplecov`
* Adding a `rake` task `:coverage` which calls `coverage:rcov` (Ruby < 1.9) or `coverage:simplecov` (Ruby >= 1.9) respectively

hence staying backward compatible and solving issue [MODULES-1206](https://tickets.puppetlabs.com/browse/MODULES-1206).

### DIFF
--- a/lib/puppetlabs_spec_helper/module_spec_helper.rb
+++ b/lib/puppetlabs_spec_helper/module_spec_helper.rb
@@ -1,4 +1,8 @@
-if ENV['COVERAGE'] == 'SimpleCov'
+case ENV['COVERAGE']
+when 'Coveralls'
+  require 'coveralls'
+  Coveralls.wear!
+when 'SimpleCov'
   require 'simplecov'
   SimpleCov.start
 end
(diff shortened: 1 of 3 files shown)


lib/puppetlabs_spec_helper/module_spec_helper.rb
```diff
@@ -1,3 +1,12 @@
+case ENV['COVERAGE']
+when 'Coveralls'
+  require 'coveralls'
+  Coveralls.wear!
```
DavidS: This needs to setup a exclude for the spec directory, like here https://github.com/puppetlabs/puppetlabs_spec_helper/pull/99/files#diff-c7beddb7111031927788e851096dc9deR38 and here https://github.com/DavidS/puppetlabs-stdlib/commit/d668dd9165699716d5706883b464401f4c0fe146#diff-93830fa29d616f7c87903d08b5b1b29aR14

PR #100, opened 2015-04-08
Updates for puppet 4

hunner:
Puppet 4 uses the environmentpath instead of modulepath and manifest
settings. We can set environmentpath to spec/ and use an environment of
'fixtures' to get catalogs.

### DIFF
--- a/lib/puppetlabs_spec_helper/module_spec_helper.rb
+++ b/lib/puppetlabs_spec_helper/module_spec_helper.rb
@@ -11,7 +11,8 @@ def verify_contents(subject, title, expected_lines)
   (content.split("\n") & expected_lines).should == expected_lines
 end
 
-fixture_path = File.expand_path(File.join(Dir.pwd, 'spec/fixtures'))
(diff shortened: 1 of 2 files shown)


lib/puppetlabs_spec_helper/module_spec_helper.rb
```diff
@@ -19,6 +20,7 @@ def verify_contents(subject, title, expected_lines)
 module_path = [module_path, env_module_path].join(File::PATH_SEPARATOR) if env_module_path
 
 RSpec.configure do |c|
+  c.environmentpath = spec_path
```
bmjen: Should this be
```
c.environmentpath = spec_path if Puppet.version.to_f >= 4.0
```

To preserve the same configuration for older puppet versions?

hunner: As it is just setting rspec settings, I'm not worried about it affecting puppet. It's up to rspec-puppet to use the settings or ignore the setting based on the puppet version.

bmjen: :+1:

PR #103, opened 2015-05-01
(MODULES-2002) Accept alternate fixtures file

op-ct:
Without this patch, the location of the fixtures YAML file is hard-coded
to be ".fixtures.yaml."  This does not accommodate modules that are
tested across environments with different module locations.

This patch introduces an environment variable FIXTURES_YML, which can
be used to specify the location of an alternate fixtures data file.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -29,7 +29,7 @@ end
 
 def fixtures(category)
   begin
-    fixtures = YAML.load_file(".fixtures.yml")["fixtures"]
+    fixtures = YAML.load_file( ENV['FIXTURES_YML'] || '.fixtures.yml' )['fixtures']
   rescue Errno::ENOENT
     return {}
   end


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -29,7 +29,7 @@ def source_dir
 
 def fixtures(category)
   begin
-    fixtures = YAML.load_file(".fixtures.yml")["fixtures"]
+    fixtures = YAML.load_file( ENV['FIXTURES_YML'] || '.fixtures.yml' )['fixtures']
```
jhoblitt: I realize that this isn't true of the other env vars this module uses but shouldn't this be namespaced?

op-ct: Should it?  Since none of the other env vars are namespaced, I'm not sure what the prefix would be.  Did you have one in mind?

trevor-vaughan: I just wanted to hop in and indicate that this would be VERY helpful on consulting engagements. And, as Chris mentioned, all of the other variables are not namespaced so there's really no precedence.

What would it take to get this pushed through?

DavidS: puppetlabs_spec_helper is scheduled for an overhaul, so even if I merge this, I don't know whether there'll be soon a release of any of this, nor if the future interface will look like this.

Other than that, I suppose this could be merged, and looking at https://github.com/puppetlabs/puppetlabs_spec_helper/pull/106 , a release might be useful anyways, if it is not too much of a hassle. @hunner?

DavidS: BTW, how to you pees feel about https://github.com/puppetlabs/puppetlabs_spec_helper/pull/107 ?

trevor-vaughan: I get that. But, considering that these are minor changes and most people have Gemfile.lock in place when testing, what is the harm?

We'll all yell if it's broken ;-)

trevor-vaughan: :+1 to #107. BUT, I want to be able to disable it.

trevor-vaughan: @DavidS: I'd also like to point out that a major overhaul of puppetlabs_spec_helper is probably going to cause issues for people who got overly creative with it and may delay the public roll out of a new version if past experience is anything to go by.

DavidS: At the same time I'd like to get rid of the need to "get overly creative with it." If you have any public examples you could point me to, I'd be grateful.

alexjfisher:
Hi.  Since it looks like there's a discussion over a few of the open PRs I thought I'd mention I'd also find this feature useful if it was merged.

DavidS:
Rebased and added docs at https://github.com/puppetlabs/puppetlabs_spec_helper/commit/495373ccc36c1741f05003c593d2a4807d2a48e8

PR #107, opened 2015-06-05
Fetch hard dependencies from metadata.json

mcanevet:


mcanevet:
Maybe we should also extend `.metadata.json` to have soft dependencies in another field so that we don't need `.fixtures.yml` anymore.
Another thing that would be great is to have a `spec_mindep_prep` and `spec_maxdep_prep` so that we can run the unit tests towards dependencies bounds using:
```shell
rake spec_mindep_prep && rake spec && rake spec_maxdep_prep && rake spec
```
Any suggestion @ryanycoleman ?

DavidS:
This is something I wanted for a long time. Just yesterday, pain could have been avoided by having something like that. What I'd really like to see though, is some way to get the set of dependencies that r10k would d/l for this module. That way, the tests would run against the set of modules that would be encountered in the wild.

tl;dr: :+1:; with possibilities for improvement.

mcanevet:
@DavidS I thought r10k ignores dependencies...

DavidS:
My impression was that it uses `puppet module` tool under the hood to d/l stuff rom the forge with all deps:
https://github.com/puppetlabs/r10k/blob/master/doc/puppetfile.mkd#forge-1

I do admit to not having actual working knowledge of the subject though. I'll see if I can get some PDX peeps on the horn.

alexjfisher:
:+1: 

@DavidS I think R10k version 2 is different in this regard.

https://github.com/puppetlabs/r10k/blob/master/CHANGELOG.mkd
> R10k can now directly install modules from the Puppet Forge, rather than shelling out to the Puppet module tool. This will allow for later optimizations like caching module downloads to speed up installing module across multiple environments.

mcanevet:
@DavidS what can I do to move forward on this topic?

DavidS:
I did some research. puppet module install does indeed download and install dependencies. r10k does not (https://github.com/puppetlabs/r10k/issues/38). @hickey has https://github.com/puppetlabs/puppetlabs_spec_helper/pull/109 up, we were discussing how to integrate metadata.json and .fixtures.yml properly there.

rndmh3ro:
I'd really like to have this feature! Is there any progress on this?

mcanevet:
@DavidS I go around my old PR and I found this one. Don't you think this is still relevant? Or maybe puppetlabs_spec_helper should use https://github.com/puppetlabs/beaker-module_install_helper so that we can have a single source of truth, the metadata.json file?

scotje:
Hey folks!

I definitely agree this is a pain point in the current flow with regards to module testing. I think the general consensus is that the `puppet module` tool CLI is deprecated so we shouldn't be building new things on top of it. So as it stands, I don't think we will merge this as it currently is implemented;  but I think there are other ways we could implement similar functionality between PDK and PSH.

I feel like the best approach at this point would be to [propose an RFC in the pdk-planning repo](https://github.com/puppetlabs/pdk-planning#pdk-rfc-process) describing at least the high level expectations of how the dependency management workflow should behave for module testing, then we can collaborate on that RFC to flesh-out the technical implementation details.

@mcanevet if you don't feel up to writing an RFC, you can just open an issue on the pdk-planning repo as a general feature request.

PR #109, opened 2015-07-03
Feature/fixture rebuild

hickey:
Makes no sense to have to maintain two separate locations for dependencies. The fixture_rebuild task will read the dependencies from metadata.json and add them as forge_modules as to the .fixture.yml file.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -289,7 +289,7 @@ task :fixture_rebuild do
     fail 'Unable to create new .fixtures.yml file'
   end
 end
-
+task :spec_prep => ['fixture_rebuild']
 
 desc "Display the list of available rake tasks"
 task :help do
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -1,6 +1,7 @@
 require 'rake'
 require 'rspec/core/rake_task'
 require 'yaml'
+require 'json'
 
 task :default => [:help]
 
@@ -133,7 +134,7 @@ task :spec_prep do
       ref = ""
     elsif opts.instance_of?(Hash)
       target = opts["target"]
-      ref = "--version #{opts['ref']}"
+      ref = "--version '#{opts['ref']}'"
     end
     next if File::exists?(target)
     unless system("puppet module install " + ref + \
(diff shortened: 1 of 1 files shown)


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -254,6 +255,42 @@ def revision(scm, target, ref)
...
+  begin
+    fixtures = YAML.load_file('.fixtures.yml')['fixtures']
+  rescue Errno::ENOENT
+    fail "Unable to load .fixtures.yml"
+  end
+  
+  begin
+    metadata = JSON.parse(File.read('metadata.json'))
+  rescue Errno::ENOENT
+    fail ".fixtures.yml does not exist"
```
DavidS: This error message is wrong

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -254,6 +255,42 @@ def revision(scm, target, ref)
...
+  rescue Errno::EACCES
+    fail "Unable to read .fixtures.yml"
```
DavidS: same here.

DavidS:
I'm not convinced that rewriting the fixtures.yml is the right way to go forward. I do like the Gemfiles' `gemspec` thing though, this could be similar, by adding a special key to the fixtures.yml that will auto-include the dependencies from metadata.json.

lib/puppetlabs_spec_helper/rake_tasks.rb
hickey: Can you clarify what you mean that error message is wrong and which one(s)? I did a moderate amount of testing with moving .fixtures.yml out of the way and setting perms/owner to validate the error messages.

hickey: Ah. sorry, I just looked at the code again and it hit me like duh!... OK fair enough....

hickey:
Hmmmm.... I got real tired trying to keep metadata.json and .fixtures.yml in sync and up to date. Hence the thought of updating .fixtures.yml with the dependency info, but unfortunately this does do things like remove extraneous info from the file (i.e. comments). I did want to make sure that nothing other than the forge_modules key was affected. 

Maybe if the forge_modules key is set to metadata.json or some other keyword, then the processing of dependencies in the metadata.json would take place. That seems reasonable.

DavidS:
Yeah, I know. "How tedious" was my first thought when I saw fixtures.yml. I really like how Gemfiles just have the simple "gemspec" declaration to include the gem's runtime dependencies into the bundle configuration (see http://bundler.io/v1.5/man/gemfile.5.html#gemspec for details).

In the .fixtures.yml, this would probably look like this:

```
---
fixtures:
  include_metadata: true
  # continue with other keys here..
```

where `include_metadata` could either be `true`, `false`, or absent (same as `false`). and setting it to `true` would imply loading the dependencies from metadata.json.

DavidS:
the puppet librarian has similar syntax in the Puppetfile: http://librarian-puppet.com -> "Simple Usage"

alexjfisher:
Being able to generate a .fixtures.yml file from an R10K puppetfile would be useful too.  The .fixtures in my role module is huge and a pain to keep in sync with the control-repo's puppetfile.
Kind Regards,
Alex

daenney:
I prefer @DavidS suggestion of `include_metata: true` and then also `include_puppetfile: true` allowing people to define where to load their dependencies from, or simply keep specifying it themselves.

logicminds:
I started working on this problem to not only fix this issue but also help with the generation of the fixtures file. https://github.com/logicminds/genfix

I think I solved most of the issues, with the exception of trying to merge metadata dependencies with fixtures. Also the genfix code requires librarian-puppet to work but your not required to use librarian-puppet in your code. So I might have to vendor librarian-puppet to make it seamless.   Please don't include this gem as its not complete yet.

DavidS:
Yesterday @rnelson0 released https://github.com/rnelson0/puppet-generate-puppetfile which potentially also has a solution for this.

david22swan:
@hickey Given the length of time that this has been left to lie fallow and the fact that it has been rebased and continued in another pr I feel that it should now be closed. Thank you for the wrkm that you have put in and I look forward to seeing more from you in the future.
Best Wishes

PR #112, opened 2015-07-27
Automatically symlink the module directory

daenney:
It's silly to require everyone to add a `source: $module_name: ${source_dir}`
to every fixtures file. This takes care of it by simply grabbing the name
of the current directory and creating the symlink stanza for you if there
isn't one.

If there's no fixtures file at all we still define the symlink
meaning that we can forego the fixtures file entirely if we do
not need to load additional modules from git/forge.

This even works when `rake` is called from subdirectories within
spec/ or any other part of the module because rake searches upwards
to find the Rakefile and executes that making `Dir.pwd` always be
reliably splittable to get the module name, unless things are not
in autoloader layout.

When this behaviour is undesired people can simply add an empty
symlinks stanza `symlinks: {}` or `symlinks: ''` instead.

I think it's far more likely that people will need to add one to
link their module than the inverse saving a lot of people some
trouble.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -27,6 +27,10 @@ def source_dir
   Dir.pwd
 end
 
+def auto_symlink
+  { Dir.pwd.split('/').last.split('-').last => '#{source_dir}' }
+end
+
 def fixtures(category)
   if File.exists?('.fixtures.yml')
     fixtures_yaml = '.fixtures.yml'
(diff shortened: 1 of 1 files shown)


daenney:
So this certainly qualifies as magic. Also if someone uses a different split character like an _ for whatever reason or a / this won't work for them so they'll still have to manually add a `symlink` stanza.

DavidS:
I'd love to hear what you think about #109 and how this will interact. I'd love to merge both on #puppethack day (Thursday)

gdubicki:
Can you please consider merging this PR without #109, @DavidS ? It looks like #109 has stalled and it would really help me to have the auto-symlink.

DavidS:
I rebased and pushed this as [8611708b3e8600d80832bc67af8c61a9c108f0d6](https://github.com/puppetlabs/puppetlabs_spec_helper/commit/8611708b3e8600d80832bc67af8c61a9c108f0d6).

PR #113, opened 2015-08-04
--recursive option for fixtures cloning from git

pondohva:
This option in .fixtures.yml allow to clone the repo with submodules into fixtures directory


domcleal:
Looks like this has been superseded by #115 / 4eeeec9, allowing any git flags to be passed through.

DavidS:
Yup, this can now be achieved by specifying `flags: --recursive`

PR #114, opened 2015-08-07
Add types in the rake tasks

roidelapluie:
As documented in https://github.com/rodjek/rspec-puppet types is also a valid directory for spec tests.


DavidS:
Thanks!

PR #115, opened 2015-08-17
Support for additional puppet and scm flags in .fixtures.yml

balous:
Hi,

I've prepared a modification that allows the user to add additional puppet and scm flags into .fixtures.yaml so that the user is able to use puppet/scm features not supported by this module.

I'd appreciate this change to be included into the mainline of this otherwise useful module. Please let me know it there's something wrong with the proposed change.

Thanks,

Petr.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -157,7 +157,11 @@ When specifying the repo source of the fixture you have a few options as to whic
    ref: 880fca52c
    ```
  * branch - used to specify the branch name you want to use ie. `branch: development`
- 
+ * flags - additional flags passed to the module installer (both puppet and scm)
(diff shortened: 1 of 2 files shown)


DavidS:
This looks very nice. Please add a few words to the README describing the changes you made, so I can merge it.

balous:
I've amended my previous commit with a README change. No it should be complete.

DavidS:
I added a small whitespace fix on top and merged. Thanks for your contribution!

balous:
Not at all, I need the fix :-). BTW when do you plan to make a next release?

Petr.

PR #117, opened 2015-10-07
(MAINT) remove environment initialisation that is already done by rsp…

DavidS:
…ec-puppet

See also https://github.com/rodjek/rspec-puppet/commit/41257b33cb1f9ade4426b044f70be511b0c89112
and https://github.com/puppetlabs/puppet/blob/423f9b3aa8a987b7e4e25a8131eaad9754c4131a/lib/puppet/test/test_helper.rb#L90-L92

### DIFF
--- a/lib/puppetlabs_spec_helper/puppet_spec_helper.rb
+++ b/lib/puppetlabs_spec_helper/puppet_spec_helper.rb
@@ -32,46 +32,24 @@ require 'puppetlabs_spec_helper/puppetlabs_spec/files'
 # You may want to hold your nose before you proceed. :)
 #
 
+# Here we attempt to load the new TestHelper API, and print a warning if we are falling back
+# to compatibility mode for older versions of puppet.
+begin
+  require 'puppet/test/test_helper'
+rescue LoadError => err
+end
 
 # This is just a utility class to allow us to isolate the various version-specific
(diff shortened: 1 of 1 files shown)


lib/puppetlabs_spec_helper/puppet_spec_helper.rb
```diff
@@ -149,14 +116,14 @@ def handle(msg)
...
-
 # And here is where we do the main rspec configuration / setup.
 RSpec.configure do |config|
+  # Some modules depend on having mocha set up for them
   config.mock_with :mocha
 
   # determine whether we can use the new API or not, and call the appropriate initializer method.
   if (defined?(Puppet::Test::TestHelper))
-    Puppet::PuppetSpecInitializer.initialize_via_testhelper(config)
+    # This case is handled by rspec-puppet since v1.0.0 (via 41257b33cb1f9ade4426b044f70be511b0c89112)
```
tphoney: do we need an if / else ? should it be an unless now

DavidS: I needed a place for that comment. Would you prefer 

```
unless ...
end
# comment here
```
?

tphoney: I understand your argument. your original code is fine.

PR #124, opened 2016-01-20
Allow injecting options for metadata-json-lint

felixb:
This allows running `metadata-json-lint` with `--no-strict-license` or `--no-fail-on-warnings`.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -2,6 +2,13 @@ require 'rake'
 require 'rspec/core/rake_task'
 require 'yaml'
 
+# optional gems
+begin
+  require 'metadata-json-lint/rake_task'
+rescue LoadError
+  # ignore
+end
+
 task :default => [:help]
 
 desc "Run spec tests on an existing fixtures directory"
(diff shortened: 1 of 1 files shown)


DavidS:
I'd rather see this replaced by using the linter's raketask https://github.com/voxpupuli/metadata-json-lint/blob/master/lib/metadata-json-lint/rake_task.rb ; then you can just set options in your Rakefile (see https://github.com/voxpupuli/metadata-json-lint/blob/master/lib/metadata_json_lint.rb#L6-L24) using your setup as you like

felixb:
We now try to require the gem and invoke the original rake task as suggested.
Setting options is working fine.

Thanks for the hint. :+1:

DavidS:
Excellent. Thanks!

garethr:
So, this broke the API for the rake tasks in a minor version bump :( The task names are basically a public API given how rakefiles work. Also this change was not mentioned in the Changelog (which also states the module follows semver).

The change from metadata to metadata_lint is backwards incompatible, and anyone not pinning to a specific version is about to find all their tests fail without local changes.

I'd recommend pushing a backwards incompatible shim, if you really want with a warning that metadata is deprecated and you should use metadata_lint.

cc. @hunner @bmjen

PR #125, opened 2016-01-26
Addition of check tasks and some formatting

HelenCampbell:
The 4 new check rake tasks include:
* symlinks - make sure we don't have any 
* test_file - Move `tests` directory to `examples` directory warning ("tests/*.pp" should be empty)
* dot_underscore - avoid adding ._* files as OS X ACLs
* git_ignore - cross-check .gitignore: there should be no files checked in that are ostensibly ignored by git

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -298,6 +298,58 @@ task :compute_dev_version do
   print new_version
 end
 
+desc "Runs all nessesary checks on a module in preparation for a release"
+task :release_checks do
+  Rake::Task[:lint].invoke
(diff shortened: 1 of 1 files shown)


DavidS:
* Awesome!
* split out the white space changes into a separate commit,
* make the tasks fail instead of warn
* use a rake [namespace](http://theadmin.org/articles/organizing-rake-using-namespaces/) to group the tasks instead of repeating it in the task name
* create a single :release_checks task to call them all at once
* add the :lint, :validate and :spec tasks to this task

Action item for later: update modules_sync to require this version of puppetlabs_spec_helper and update travis.yml to call this new task

HelenCampbell:
@DavidS Think I got everything there, have a look when you're free :)

PR #126, opened 2016-01-29
parallel fixture downloads

logicminds:
* this will parallelize fixture downloads for git
    which is at a minimum 520% faster than using 1 thread.
  * thread amount can be configured by using MAX_FIXTURE_THREAD_COUNT
    environment variable.
  * refactors the fixtures repositories retrieval so that it does not
    parse yaml more than once.
  * added a logger in order to debug threads and such if needed, which
    can be enabled by setting ENABLE_LOGGER=true


logicminds:
My roles and profiles modules have 100+ fixtures in them which all rely on git. Its painful to test these so I thought I would make it faster.   I also needed a way to limit the number of threads because my gitlab server can only handle around 50 before ssh starts denying sessions so I added a environment variable and set the default to 10.  I have also found it much quicker (few seconds) if I add this to my ssh config.

```
Host github.com
  ControlMaster auto
  ControlPath ~/.ssh/ssh-%r@%h:%p
  ControlPersist yes

``` 

Below are some results which are pretty staggering. Took me about an hour to code this up.  Wish I would have done this sooner. 

```
rake spec_prep  3.90s user 2.71s system 18% cpu 35.070 total  # 10 threads  520% faster
rake spec_prep  3.87s user 2.55s system 26% cpu 24.558 total  # 15 threads  755% faster
rake spec_prep  3.80s user 2.58s system 35% cpu 17.885 total  # 20 threads  1039% faster 
rake spec_prep  3.91s user 2.66s system 43% cpu 14.992 total  # 25 threads  1233% faster
rake spec_prep  3.97s user 2.72s system 57% cpu 11.596 total  # 50 threads  1608% faster
rake spec_prep  3.82s user 2.47s system 3% cpu 3:05.64 total  # 1 thread (current way)
```

DavidS:
delicious! 

the travis boo boo doesn't look related to your changes, but it does hide a proper test run of your code, and I'll be on the road next week.

logicminds:
do we have any idea what that error is?

logicminds:
@DavidS Does the repo even get tested?  No travis file, weak Rakefile, tests are broken in master branch, gem spec/Gemfile needs puppet as a development dependency ...

Am I doing something wrong.  I updated a few more files to correct these things, but tests are still failing unrelated to change.

logicminds:
Turns out there was a bug in bundler causing the travis related error.  The travis file I added also updates bundler so the tests will actually run now.  However, many of the tests are broken even before I added my stuff.

logicminds:
This also combines PR #127 which fixes all the pre-existing issues with the tests.

Gemfile
```diff
@@ -1,7 +1,22 @@
...
+  gem 'yard'
+  gem 'pry'
+  gem 'jeweler'
+  gem "puppet", ENV['PUPPET_VERSION'] || '~> 3.8.3'
+end
+
+group :test do
+end
+
+# I don't know the purpose of this below
```
hunner: This was so someone could add more gems by making a .local file... I'd probably just delete this stuff

spec/spec_helper.rb
```diff
@@ -0,0 +1 @@
+require 'pry'
```
hunner: I'd rather not add this

DEVELOPMENT.markdown
```diff
@@ -0,0 +1,29 @@
+# DEVELOPMENT NOTES
```
hunner: This file needs removed; duplicate

DEVELOPMENT.md
```diff
@@ -0,0 +1,29 @@
+# DEVELOPMENT NOTES
```
hunner: This whole file should probably be turned into a `CONTRIBUTING.md` rather than `DEVELOPMENT.md`

lib/puppetlabs_spec_helper/version.rb
```diff
@@ -1,5 +1,3 @@
 module PuppetlabsSpecHelper
-  module Version
-    STRING = '1.0.1'
-  end
+  VERSION = '1.0.1'
```
hunner: Could we leave PuppetlabsSpecHelper::Version::STRING just in case someone is using it?

puppetlabs_spec_helper.gemspec
```diff
@@ -1,25 +1,86 @@
...
 
 Gem::Specification.new do |s|
-  s.name        = "puppetlabs_spec_helper"
-  s.version     = PuppetlabsSpecHelper::Version::STRING
-  s.authors     = ["Puppet Labs"]
-  s.email       = ["modules-dept@puppetlabs.com"]
-  s.homepage    = "http://github.com/puppetlabs/puppetlabs_spec_helper"
-  s.summary     = "Standard tasks and configuration for module spec tests"
+  s.name = "puppetlabs_spec_helper"
+  s.version = "1.0.1"
```
hunner: Should this pull from PuppetlabsSpecHelper::VERSION?

logicminds: I ended up just putting back the old Version::STRING so it would not look so weird.

DEVELOPMENT.md
logicminds: moved file to CONTRIBUTING.md

logicminds:
Please note that when building the gemspec with rake, the gem file will pull in whatever version of puppet specified as a development dependency. Wondering if we should hard code the puppet version in the gem file since its currently using the environment variable or defaulting to 3.8.3.  Previously puppet was not specified as a development dependency and tests would fail as a result which is why I added it. 

```
group :development do
  gem 'yard'
  gem 'pry'
  gem 'jeweler'
  gem "puppet", ENV['PUPPET_VERSION'] || '~> 3.8.3'    <<<<<--------
end
```

@hunner 

I have addressed all the points above.  Please see the CONTRIBUTING doc for info on how to release.

hunner:
Could you squash this into logical commits? Thanks!

PR #127, opened 2016-01-29
Fix testing

logicminds:
Testing is completely broken and probably has been for a while.
While this doesn't fix any tests currently, it at least brings the gem up to speed with some new rake tasks. 

I was going to fix all the tests but it seems to be more involved than I had planned.


logicminds:
This brings this gem back into a working testable state.  Previously most of tests were failing due to issues with changing puppet versions, bundler and ruby.  Without this the tests are broken for everyone going forward.

Note:  test coverage is pretty low, so I assume manual integration testing has been used instead.

hunner:
Closing as it is part of #126

PR #131, opened 2016-02-29
Re-add the missing metadata task with a depreciation warning

DavidS:
Ref: https://github.com/puppetlabs/puppetlabs_spec_helper/pull/124#issuecomment-189475146


garethr:
+1. Thanks for the quick fix.

PR #136, opened 2016-03-06
[MODULES-3128] Ignore symlinks inside .git when running check:symlinks rake task

fvoges:
Fix for MODULES-3128

Before this change rake check:symlinks would search for symlinks
anywhere within the project directory. This meant that using symlinks to
install git hooks would trigget an error (false possitive).

This change modifies the check:symlinks task to exclude the .git
directory from the check:symlinks search


DavidS:
Thanks for the patch!

PR #140, opened 2016-03-12
Add RuboCop to both gem and module rake tasks

leoarnold:
This project has been describe as [dogfood](https://github.com/puppetlabs/puppet/blob/d2199af906fa5ca19ec0c37c82794934ea4fb4b5/spec/spec_helper.rb#L2) and its code is somewhat unclean and not easily comprehensible.

As a first countermeasure I suggest to add RuboCop to the project as well as all modules using this gem.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 
 gem 'rake'
 gem 'rspec-puppet'
+gem 'rubocop'
 gem 'puppet-lint'
 gem 'puppet-syntax'
 gem 'mocha'
--- a/Rakefile
+++ b/Rakefile
@@ -46,5 +46,8 @@ end
 
 task :default => :spec
 
+require 'rubocop/rake_task'
+RuboCop::RakeTask.new
+
 require 'yard'
 YARD::Rake::YardocTask.new
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -1,5 +1,6 @@
 require 'rake'
 require 'rspec/core/rake_task'
+require 'rubocop/rake_task'
 require 'yaml'
 
 # optional gems
@@ -303,6 +304,8 @@ desc "Clean a built module package"
 task :clean do
   FileUtils.rm_rf("pkg/")
 end
+
+RuboCop::RakeTask.new
 
 require 'puppet-lint/tasks/puppet-lint'
 # Must clear as it will not override the existing puppet-lint rake task since we require to import for
--- a/puppetlabs_spec_helper.gemspec
+++ b/puppetlabs_spec_helper.gemspec
@@ -55,6 +55,7 @@ Gem::Specification.new do |s|
     if Gem::Version.new(Gem::VERSION) >= Gem::Version.new('1.2.0') then
       s.add_runtime_dependency(%q<rake>, [">= 0"])
       s.add_runtime_dependency(%q<rspec-puppet>, [">= 0"])
+      s.add_runtime_dependency(%q<rubocop>, [">= 0"])
       s.add_runtime_dependency(%q<puppet-lint>, [">= 0"])
       s.add_runtime_dependency(%q<puppet-syntax>, [">= 0"])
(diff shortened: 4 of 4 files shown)


alexjfisher:
I think this means puppetlabs_spec_helper is now not installable on ruby 1.8 ??

leoarnold:
@alexjfisher Ruby 1.8 was [retired in Mid 2013](https://www.ruby-lang.org/en/news/2013/06/30/we-retire-1-8-7/) and is [not supported by Puppet 4](https://docs.puppet.com/puppet/4.0/reference/system_requirements.html#ruby).

alexjfisher:
Hi!
Yep, I know.  But it's the default system ruby on RHEL6 and support for that doesn't end for years.

Because RHEL6 is so widely deployed, I doubt I'm the only puppet developer who has to write/maintain modules that are compatible with ruby 1.8, (I also make sure they work with puppet 4 and its AIO ruby).

Would it be possible to make rubocop a dependency only for ruby >= 1.9 and disable the new rake task if it's not installed?

Either way, if you really are going to remove ruby 1.8 support, it might be nice to release 1.1.2 with a deprecation warning or something.  It's never nice to wake up one morning and find all of your module's CI builds are failing :)

leoarnold:
@alexjfisher You can still develop modules for Ruby 1.8. RuboCop can be configured to enforce Ruby 1.8 syntax. Only your development system needs to run a contemporary Ruby.

Also, Puppet Collection 1 / Puppet 4 Agent for EL6 contains its own Ruby 2.1.0 distribution, so everything should be good.

alexjfisher:
But I can no longer run spec tests that test my functions and providers work with ruby 1.8.

I could of course pin puppetlabs_spec_helper to 1.1.1.  That'd be a shame though, as I only discovered ruby 1.8 no longer worked when working on a simple patch I was planning on submitting.

I have started to prepare for puppet 4, but migrating is still some way off.  I use the foreman, and it doesn't *yet* fully support puppet 4.

leoarnold:
@alexjfisher This PR should **not** have affected the spec testing mechanism at all. Please provide some error messages for clarification.

alexjfisher:
oh right.  Not at my dev machine at the moment, but basically, you can't do a bundle install any more.
IIRC It fails when it tries to install 'rainbow' (which requires ruby 1.9.2).  rainbow is a dependency of rubocop.  I stepped through each commit since 1.1.1 until I hit this one.

I can attach a gist in the morning (BST) if it helps.

leoarnold:
I see. But why would you want to do a `bundle install` including development dependencies on a production machine?

In your `Gemfile`, you could put the spec helper in a `group :development`. On production machines, you would then just run `bundle install --without development`.

alexjfisher:
Sorry, now I'm really confused...

I don't run ```bundle install``` on production machines.  I use my dev machine (or CI environment) to run rspec-puppet tests (including function tests where I'm testing ruby code, not puppet manifests).
eg something like...
```
cd my_module
rvm use 1.8.7
bundle install
PUPPET_GEM_VERSION='~>3.8.0' bundle exec rake lint validate spec
```
Then repeat for other versions of ruby, puppet, future parser and strict variables settings etc.  In modules that don't contain a ```lib``` directory, not testing against ruby 1.8 doesn't matter, but in the modules that do, I think it's still important to quite a few people.

leoarnold:
Ah, that's quite a nested case I did not consider. Sorry about that. I'll revise this and submit a new PR on the weekend.

alexjfisher:
@leoarnold Thanks!

I think it's actually quite a common scenario.  60% of these [most popular puppet approved modules](https://forge.puppet.com/modules?utf-8=%E2%9C%93&sort=downloads&page=1&endorsements=approved) explicitly test against ruby 1.8.

https://github.com/stankevich/puppet-python/blob/master/.travis.yml#L7
https://github.com/garethr/garethr-docker/blob/master/.travis.yml#L7
https://github.com/hercules-team/augeasproviders_core/blob/master/.travis.yml#L3
https://github.com/garethr/garethr-erlang/blob/master/.travis.yml#L7
https://github.com/sensu/sensu-puppet/blob/master/.travis.yml#L10
https://github.com/saz/puppet-rsyslog/blob/master/.travis.yml#L12

PR #141, opened 2016-03-12
Display list of rake tasks by default

leoarnold:
The rake tasks provided by this gem display the list of rake tasks by default.
The Rakefile of this gem should do the same to avoid confusion.


hunner:
Nice. Could you rebase this?

leoarnold:
Rebased.

PR #142, opened 2016-03-12
Run spec tests against Puppet 4

leoarnold:


README.markdown
```diff
@@ -1,6 +1,8 @@
 Puppet Labs Spec Helper
 =======================
 
+[![Build Status](https://travis-ci.org/puppetlabs/puppetlabs_spec_helper.svg)](https://travis-ci.org/puppetlabs/puppetlabs_spec_helper)
```
hunner: Could you actually not add badges to the readme? They end up showing up in all releases but only reflect the status of the master branch and can be misleading.

leoarnold:
Travis CI Badge removed. Commit rebased.

leoarnold:
The build error seems to be due to an update in the Puppet gem. Please wait while I investigate it.

leoarnold:
The function `split()` moved from `puppet/parser/functions` to `puppet/functions` with [this commit](https://github.com/puppetlabs/puppet/commit/65a6806a6db3c7797b1a5981d64e99c66fdcd904). Since the test is about function testing (and not `split()` in particular) I just replaced it with a test for `tag()`, which (still) is in `puppet/parser/functions`. Now rebased and passing.

PR #143, opened 2016-04-02
Add more beaker actions

dmitryilyin:
* Renamed beaker_nodes to beaker:sets
  Old name aliaced to the new one.
* Add beaker:ssh action to login to the node
* For every node set add beaker:$set action to
  run the tests for this set and beaker:ssh:$set
  to login to this test node

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ Gemfile.lock
 pkg/
 .*.sw*
 .bundle/
+.idea
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -54,6 +54,39 @@ def repositories
   @repositories
 end
 
+# get the array of Beaker set names
+# @return [Array<String>]
+def beaker_node_sets
+  return @beaker_nodes if @beaker_nodes
+  @beaker_nodes = Dir['spec/acceptance/nodesets/*.yml'].sort.map do |node_set|
+    node_set.slice!('.yml')
+    File.basename(node_set)
+  end
+end
+
+# Use "vagrant ssh" to login to the given node in the node set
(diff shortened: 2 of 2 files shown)


DavidS:
@zreichert , @gregohardy : please review interface here

dmitryilyin:
ping

DavidS:
This overlaps with [rototiller](https://github.com/puppetlabs/rototiller), a replacement for this part of PSH, perhaps @er0ck can direct some attention on this?

dmitryilyin:
Hmm, this **rototiller** looks very nice and promising.
Looks like it's still in an early stage of development but **puppetlabs_spec_helper** is being used pretty much everywhere.
Perhaps, I could help with **rototiller** instead?

dmitryilyin:
Maybe **puppetlabs_spec_helper** can be refactored to use **rototiller** for rake tasks generation?

DavidS:
@dmitryilyin using rototiller as new base for PSH would be the plan, once a stable release is out. I can't make any promises at the current time though, and all rototiller folks are on PDX timezone, so asleep for another five hours.

er0ck:
rototiller's user stories, as of now, do not exactly cover what it appears these tasks are for.  They will make writing these tasks easier and more uniform, sometime soon.  The first release of rototiller, which we expect very soon, is primarily geared at environment variable tracking and messaging and tasks with commands and their flags (with environment variable linking/override)
We may want to investigate the value of centralizing a task for things like `beaker:nodes` as above.

dmitryilyin:
Maybe we can merge this PR and refactor env variable related stuff to rototiller later?

er0ck:
@DavidS i'd suggest what @dmitryilyin said, solely with regards to any future Rototiller related execution.

DavidS:
Thanks everyone for their input!

PR #147, opened 2016-04-18
Remove RuboCop to fix spec testing on Ruby 1.8

leoarnold:
Ruby 1.8 is deprecated, but is still supported by PE 3.8 which is still maintained.
As @alexjfisher pointed out in the discussion following up the merge of PR #140,
RuboCop (or at least one of its dependencies) is not compatible with Ruby 1.8.
Hence spec testing on Ruby 1.8 will break because `bundle install` fails.

Since the GemSpec is generated with Jeweler which does not support different
gem dependencies for different Ruby versions, I suggest remove RuboCop for the moment.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,6 @@ source 'https://rubygems.org'
 
 gem 'rake'
 gem 'rspec-puppet'
-gem 'rubocop'
 gem 'puppet-lint'
 gem 'puppet-syntax'
 gem 'mocha'
--- a/Rakefile
+++ b/Rakefile
@@ -46,8 +46,5 @@ end
 
 task :default => :spec
 
-require 'rubocop/rake_task'
-RuboCop::RakeTask.new
-
 require 'yard'
 YARD::Rake::YardocTask.new
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -1,6 +1,5 @@
 require 'rake'
 require 'rspec/core/rake_task'
-require 'rubocop/rake_task'
 require 'yaml'
 
 # optional gems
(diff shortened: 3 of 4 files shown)


hunner:
Would it be better to pull out the jeweler stuff and go back to the manual-ish way of publishing gems and then do per-ruby gemsets?

leoarnold:
@hunner I guess so, since this project will always have to support a range of Puppet versions which in turn each support a range of Ruby versions, so there's no hope to ever have a one-gemspec-fits-all.

hunner:
It looks like PE 3.7 or later (at least) ship with ruby 1.9 or later: https://docs.puppet.com/pe/latest/overview_version_table.html so I don't think 1.8 is a concern there.

I test using rbenv or puppet-agent, so don't have to deal with ruby 1.8 on rhel 6 any more (thank goodness). And this also gives better results by matching the versions of ruby that ship in production puppet environments.

But as to @alexjfisher 's comments from the other thread, we fix 1.8 compatibility and perform another 1.x release. Maybe then we can officially drop 1.8 support in this gem and release 2.0.0 and go from there?

DavidS:
Looking over my experiences with gems over the last year, "nobody" bothers to bump a major just for dropping off a ruby version "nobody" cares about.

Just saying.

hunner:
And I would like for us to be the exception to that lazy & irritating pattern of "drop a ruby compatibility on a minor/bugfix release." Just because everyone else does it doesn't make it right. :)

alexjfisher:
> Looking over my experiences with gems over the last year, "nobody" bothers to bump a major just for dropping off a ruby version "nobody" cares about.

@DavidS In this case, could it be that quite a few people will care though?  When I checked through some of the most popular forge modules, it certainly didn't seem that way from the `.travis.yml` files.

I didn't go out of my way to find 6 examples in https://github.com/puppetlabs/puppetlabs_spec_helper/pull/140#issuecomment-209816283
That really is from page 1 of all puppet approved modules sorted by downloads.

leoarnold:
@hunner I agree. I leave it up to you whether to close this PR or keep it open to attract further discussion.

PR #151, opened 2016-05-25
Create the fixtures directory before cloning repos

DavidS:
git has a race condition creating parent directories, that would lead to aborted clone operations:

Cloning into 'spec/fixtures/modules/staging'...
fatal: could not create leading directories of 'spec/fixtures/modules/stdlib': File exists
rake aborted!
Failed to clone git repository https://github.com/puppetlabs/puppetlabs-stdlib into spec/fixtures/modules/stdlib


alexjfisher:
:+1:

PR #144, opened 2016-04-04
add parallel_spec rake task

jlambert121:

PR #152, opened 2016-06-12
(MODULES-3456)  Adds feature :parallel_spec

alexharv074:
As described in https://rnelson0.com/2016/03/18/parallelized-rspec-tests/ and http://razorconsulting.com.au/parallelising-rspec-puppet.html.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,21 @@ Host github.com
 
 Note: parallel downloads is only available for repositories and not forge modules.
 
+### Parallel tests
+It is also possible to use the `parallel_tests` Gem via the `:parallel_spec` Rake task to run rspec commands in parallel on groups of spec files.
+
+Use of parallelization at this level can result in large performance benefits when the Rspec examples tend to cause a number of large, CPU-intensive catalog compilations to occur.  An example of where this might be the case is in a complex module with a lot of tests or a control repo with many hosts.
+
+Be aware however that in other circumstances this parallelization can result in the tests actually taking longer to run.  The best thing to do is time `rspec spec` and `rspec parallel_spec` and use the parallelization only when there is a clear benefit.
+
+To enable this feature, add the `parallel_tests` Gem to your project's Gemfile:
+
+    gem 'parallel_tests'
+
+And then to run spec tests in parallel:
+
+    $ rake parallel_spec
+
 Issues
 ======
 
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -14,11 +14,12 @@ end
 
 task :default => [:help]
 
+pattern = 'spec/{aliases,classes,defines,unit,functions,hosts,integration,types}/**/*_spec.rb'
+
 desc "Run spec tests on an existing fixtures directory"
 RSpec::Core::RakeTask.new(:spec_standalone) do |t|
   t.rspec_opts = ['--color']
 
-  pattern = 'spec/{aliases,classes,defines,unit,functions,hosts,integration,types}/**/*_spec.rb'
   if ENV['CI_NODE_TOTAL'] && ENV['CI_NODE_INDEX']
     ci_total = ENV['CI_NODE_TOTAL'].to_i
     ci_index = ENV['CI_NODE_INDEX'].to_i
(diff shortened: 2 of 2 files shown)


alexharv074:
Ping @petems @rnelson0

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -324,6 +324,22 @@ def max_thread_limit
...
+  Dir.glob(pattern).each { |f| args << f }
+
+  desc "Parallel spec tests"
+  task :parallel_spec do
+    Rake::Task[:spec_prep].invoke
+    ParallelTests::CLI.new.run(args)
+    Rake::Task[:spec_clean].invoke
+  end
+rescue LoadError
+  # ignore
```
petems: Maybe put an alert here to install the gem? But I see through the rest of the code we do a lot of empty rescues, so ¯\_(ツ)_/¯

alexharv074: I think the empty rescues are intentional, otherwise I might as well make `parallel_tests` a dependency in the gemspec.  I've done it this way so that people aren't forced to download that Gem if they don't want to use it.

petems: Ah yep, makes sense 👍

README.md
```diff
@@ -61,6 +61,17 @@ Host github.com
 
 Note: parallel downloads is only available for repositories and not forge modules.
 
+### Parallel tests
+It is also possible to use the `parallel_tests` Gem via the `:parallel_spec` task.  Use of parallelization at this level can result in large performance benefits when a system has a number of CPUs and the Rspec examples involve many full catalog compilations, such as in a control repo that compiles catalogs for every host for example.
```
rnelson0: I suggest adding a warning that it may not work in all situations, such as "...but may actually result in slowdowns with small test sets (less than 100 tests)."

petems: Good idea, a caveat to say "This is most efficent with modules with a large number of spec files, rather than spec test count"

There is a way of bucketing all the tests, and running them in parralel so it doesn't matter how many files there are, but it's a lot harder because of dependancy management and making sure all the tests can run in parralel...

alexharv074:
Upated per suggestions.

alexharv074:
Note that @jlambert121  had raised a similar PR for this back in April that hasn't been merged and I didn't see it.  https://github.com/puppetlabs/puppetlabs_spec_helper/pull/144

lib/puppetlabs_spec_helper/rake_tasks.rb
DavidS: In the case that the gem is not installed, I'd prefer an empty task that reminds the user that the gem is not installed and causes a failure. That way misconfigured CI environments fail consistently and helpfully.

alexharv074:
@DavidS   updated according to your suggestion.

DavidS:
Will a `warn` fail the task?

alexharv074:
@DavidS ah, sorry, try now.

DavidS:
I'll add it to my list of things to test.

alexharv074:
Sure thing, and any chance we could get this thing released soon?

cyberious:
I have been running this now for about a week without any issues other than my machine goes hot due to heavy usage on our 1k+ tests for profiles

cyberious:
I am actually getting issues with fixtures stubbing, i.e. using facts I am getting `Provider must have features 'versionable' to set 'ensure' to '2.6.1-2'` but no errors when running non parallel tests.

cyberious:
It appears that it breaks them differently based on `context` or `describe` Here is the example where it breaks
``````
RSpec.decribe 'my_class' do
    on_supported_os.each {  |os, facts|
        describe os do
          let(:facts){ facts.merge( { :datacenter => 'my_center' }) }
        end
     }
end
```
```
RSpec.decribe 'my_class' do
    on_supported_os.each {  |os, facts|
        context os do
          let(:facts){ facts.merge( { :datacenter => 'my_center' }) }
        end
     }
end
```

alexharv074:
@cyberious Can you show your exact setup somehow somewhere?  Fixtures stubbing works fine for me.

alexharv074:
Ok, rebased.  I didn't want the Gem in dependencies because I believe only a handful of users will need this feature, and I didn't want to force them to install parallel_tests.

DavidS:
Ok, that doesn't have to be the final decision anyways. Thanks for your work!

PR #153, opened 2016-06-22
(maint) Fix default disable checks for puppet-lint 2.0.0

glennsarti:
There was a breaking change in puppet-lint 2.0.0 which renamed the
line length check from 80 to 140 chars.  This commit updates the gem
to include this newly named check.


jrwesolo:
Nice! I've been looking forward to a new release of puppet-lint. If this is the case, should version constraint be added for puppet-lint gem >= 2.0.0?

glennsarti:
No need.  You can add disable rules that don't exist i.e. I added disable rules for both 80 and 140 so it will apply to any puppet-lint version.

jrwesolo:
Ah nice. Thanks for the explanation. That make sense now!

ghoneycutt:
+1

PR #158, opened 2016-07-14
WIP -- Break out fixtures and beaker rake tasks into sperate files

cyberious:
- More changes coming but this is the start of a refacter and enhancement


DavidS:
That looks like a good cleanup. I heard @pinkypie was looking into a similar project on the beaker side. It would be great if we could get everyone on the same page here.

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -1,4 +1,6 @@
 require 'rake'
+require 'beaker/rake_tasks'
+require 'fixtures/rake_tasks'
```
DavidS: shouldn't these both be `puppetlabs_spec_helper/...` ?

cyberious: It is a relative base and should not require the full namespace.

eputnam:
@cyberious are you planning to continue this work?

DavidS:
Superseded by #238

PR #161, opened 2016-08-10
Add user-experience to the rubocop task

DavidS:


bastelfreak:
So this will kill support for ruby <2.2.0. I really like that, but I don't think that this is want you want?

DavidS:
err.. because rubocop-rspec is only available later? that would suck.

bastelfreak:
The latest 1.6 release requires ruby2.2.0. At voxpupuli we only run rubocop on ruby2.3.1 with a few hacks in our Gemfile.

bastelfreak:
Our configs:
https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/Rakefile#L6-L13
https://github.com/voxpupuli/modulesync_config/blob/master/config_defaults.yml#L45-L48

example result:
https://github.com/voxpupuli/puppet-zabbix/blob/master/Gemfile#L29

DavidS:
This needs to be solved at the module level then, for the forseeable future :-(

DavidS:
or rather, I'll just take out the gem dependency

DavidS:
See #162 for fixed PR.

Thanks github.

PR #165, opened 2016-08-23
Update puppet-lint and puppet-syntax default ignore paths

DavidS:
These changes are to align the defaults more with what's currently in use
in the field, and improve puppet4 compatibility.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -257,7 +257,7 @@ task :spec_prep do
       opts[:thread] = Thread.new do
         clone_repo(scm, remote, target, subdir, ref, branch, flags)
         revision(scm, target, ref) if ref
-	remove_subdirectory(target, subdir) if subdir
+        remove_subdirectory(target, subdir) if subdir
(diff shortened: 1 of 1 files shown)


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -429,14 +429,24 @@ def max_thread_limit
...
+  ]
 end
 
 require 'puppet-syntax/tasks/puppet-syntax'
 PuppetSyntax.exclude_paths ||= []
 PuppetSyntax.exclude_paths << "spec/fixtures/**/*"
 PuppetSyntax.exclude_paths << "pkg/**/*"
 PuppetSyntax.exclude_paths << "vendor/**/*"
+if Puppet.version.to_f < 4.0
+  PuppetSyntax.exclude_paths << "types/**/*"
```
bastelfreak: stupid question, why is this needed?

DavidS: with puppet<4 puppet syntax will choke on type declarations.

bastelfreak:
LGTM!

PR #166, opened 2016-08-24
remove jeweler usage

logicminds:
* jeweler is dead and bloated.  Bundler has replaced most of the
    features using standard rake tasks, although not all of them.
  * This removes the jeweler library and updates the gemspec and
     rakefile to use bundler/gem_tasks instead.
  * This also adds a new rake task called `show_gemspec` for when
    you need to see what the gem build command creates.

PR #167, opened 2016-08-24
Release prep for 1.2.1

DavidS:


bastelfreak:
LGTM!

DavidS:
@bastelfreak but does it install on 1.9 ?

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -556,12 +555,15 @@ def max_thread_limit
   system("rake -T")
 end
 
-RuboCop::RakeTask.new(:rubocop) do |task|
-  # These make the rubocop experience maybe slightly less terrible
-  task.options = ['-D', '-S', '-E']
-end
-
-RuboCop::RakeTask.new("rubocop:auto_correct") do |task|
```
domcleal: Did you mean to delete this?

DavidS: yeah. rubocop rake tasks automatically add a :auto_correct task using the base task's options.

domcleal: Ah yes, I see it in rubocop now. Looks good!

domcleal:
I've tested this against theforeman/foreman locally on Ruby 1.9.3 and it at least installs a compatible set of gems and begins to run `rake spec` successfully, without any problems introduced by rubocop deps.

alexjfisher:
Works for me on ruby 1.9.3 too.

logicminds:
Bundler uses different tag names (with v) than what this gem uses.   You may want to merge the rakefile from #166 as it had some custom tasks around that workflow and gemspec preview task.

DavidS:
I don't see a reason for *not* using the current tooling's defaults. The
tags are not signed anyways.

PR #168, opened 2016-08-26
Only set strict_variables setting when required

domcleal:
Older Puppet versions (< 3.5) fail when trying to disable the
strict_variables setting that doesn't exist, preventing any tests from
running.

    ArgumentError:
      Attempt to assign a value to unknown configuration parameter :strict_variables
    # puppet-3.4.3/lib/puppet/settings.rb:786:in `set_value'
    # puppet-3.4.3/lib/puppet/settings.rb:109:in `[]='
    # puppetlabs_spec_helper-1.2.1/lib/puppetlabs_spec_helper/module_spec_helper.rb:35:in `block (2 levels) in <top (required)>'


DavidS:
... and released!

PR #171, opened 2016-12-13
(maint) Module install - ensure paths use /

ferventcoder:
Due to https://tickets.puppetlabs.com/browse/PUP-4884, modules
may fail to install on Windows. It's also somewhat hard to detect
this is occurring. Grab the full path every time for install and
ensure that the paths use forward slashes only.

cc @puppetlabs/modules

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -292,11 +292,17 @@ task :spec_prep do
     end
     next if File::exists?(target)
 
+    # https://tickets.puppetlabs.com/browse/PUP-4884
+    working_dir = File.expand_path('spec/fixtures/work-dir')
+    working_dir = working_dir.gsub(/\\/, '/')
+    target_dir = File.expand_path('spec/fixtures/modules')
+    target_dir = target_dir.gsub(/\\/, '/')
+
     command = "puppet module install" + ref + flags + \
       " --ignore-dependencies" \
       " --force" \
-      " --module_working_dir spec/fixtures/module-working-dir" \
-      " --target-dir spec/fixtures/modules #{remote}"
+      " --module_working_dir #{working_dir}" \
+      " --target-dir #{target_dir} #{remote}"
 
     unless system(command)
       fail "Failed to install module #{remote} to #{target}"
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -292,11 +292,11 @@ task :spec_prep do
     end
     next if File::exists?(target)
 
-    # https://tickets.puppetlabs.com/browse/PUP-4884
+    # The problem with the relative path is that PMT doesn't expand the path properly and so passing in a relative path here
+    # becomes something like C:\somewhere\backslashes/spec/fixtures/work-dir on Windows, and then PMT barfs itself.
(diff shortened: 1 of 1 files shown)


ferventcoder:
> Do you have a failing test log somewhere, or is this purely pro-active defensiveness?

I don't usually provide PRs for things that are just being proactive. Yes this fixes the issue. And like I said, it's maybe a bit of overkill because the gsubs probably are not required when we use expand_path.

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -292,11 +292,17 @@ def max_thread_limit
     end
     next if File::exists?(target)
 
+    # https://tickets.puppetlabs.com/browse/PUP-4884
+    working_dir = File.expand_path('spec/fixtures/work-dir')
+    working_dir = working_dir.gsub(/\\/, '/')
+    target_dir = File.expand_path('spec/fixtures/modules')
+    target_dir = target_dir.gsub(/\\/, '/')
```
ferventcoder: @DavidS let's chat here. I mentioned that the gsub bits are maybe a bit of overkill and probably unnecessary. This is fixing the issue with PUP-4884 in that it doesn't switch the paths to forward slashes by switching the paths to forward slashes here. I don't have a mistrust for File.expand_path, but maybe you didn't quite catch my humor here.

ferventcoder:
This issue is why windows modules can't modulesync up to the latest puppetlabs_spec_helper and are pinned to a particular version. It may not be the only issue, but this is a start down the correct path.

ferventcoder:
Or put another way, this causes PSH to fail on Windows every time.

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -292,11 +292,17 @@ def max_thread_limit
...
+
     command = "puppet module install" + ref + flags + \
       " --ignore-dependencies" \
       " --force" \
-      " --module_working_dir spec/fixtures/module-working-dir" \
-      " --target-dir spec/fixtures/modules #{remote}"
+      " --module_working_dir #{working_dir}" \
+      " --target-dir #{target_dir} #{remote}"
```
ferventcoder: I should have mentioned it is not super easy to see what is going on, that is why this took longer to find. This change is going from specifying a relative path to specifying a full path. The problem with the relative path is that PUP-4884 doesn't expand the path properly and so it becomes something like `C:\somewhere/spec/fixtures/work-dir`, and then it fails.

DavidS: ah! this was the piece I missed. Apologies! :man_facepalming: 

If you could put the "The problem with the relative path..." sentence into a comment in the code, that would go a long way towards making it clear for future-us to understand what's going on here.

ferventcoder:
`\o/`

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -292,11 +292,17 @@ def max_thread_limit
     end
     next if File::exists?(target)
 
+    # The problem with the relative path is that PMT doesn't expand the path properly and so passing in a relative path here
+    # becomes something like C:\somewhere\backslashes/spec/fixtures/work-dir on Windows, and then PMT barfs itself.
```
ferventcoder: 👍

ferventcoder:
Thanks for following that up!

PR #172, opened 2017-01-06
(FM-5989) add i18n tools

eputnam:
Part of i18n and l10n of modules will be making the correct tools
available to every module. This commit:
- Adds gettext-setup gem dependency ~> 0.13
- Adds and initializes rake tasks from the gem, depdendent on the
  presence of a "locales" directory.


DavidS:
How does this interact with https://github.com/puppetlabs/modulesync_configs/pull/119/files ?

eputnam:
A good question. It doesn't look like the fast_gettext dependency should be an issue just now, but perhaps we should consider changing to ~> 1.1.0 for newer ruby versions in modulesync_configs.
https://github.com/puppetlabs/gettext-setup-gem/blob/master/gettext-setup.gemspec#L20

PR #174, opened 2017-02-02
Add type_aliases directory

domcleal:
Next version of rspec-puppet will support type alias tests.


DavidS:
cc @rodjek for laughs.

DavidS:
and by "laughs", I mean weeping.

PR #179, opened 2017-02-08
(maint) Only output gettext-setup gem errors if rake is verbose

glennsarti:
Previously loading the rake task would also generate text output if the
the gettext-setup was not loaded or could not be configured.  However this
breaks the compute-dev-version rake task as it should output a single line of
the module version.  This commit confines the helpful messages to only be output
if `--verbose` is specified on the Rake command line.  By default verbose is not
set.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -23,10 +23,10 @@ begin
   if File.exist? locales_dir
     GettextSetup.initialize(locales_dir)
   else
-    puts "No 'locales' directory found in #{File.dirname(__FILE__)}, skipping gettext initialization"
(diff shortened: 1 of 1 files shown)


glennsarti:
Example output:
``` powershell
C:\Source\puppetlabs-chocolatey [master ≡ +0 ~1 -0 !]> bundle exec rake compute_dev_version
2.0.1-5631d7f1
C:\Source\puppetlabs-chocolatey [master ≡ +0 ~1 -0 !]> bundle exec rake compute_dev_version --verbose
No 'locales' directory found in C:/Source/puppetlabs-chocolatey/.bundle/windows/ruby/2.3.0/gems/puppetlabs_spec_helper-2.0.0/lib/puppetlabs_spec_helper, skipping gettext initialization
2.0.1-5631d7f1
C:\Source\puppetlabs-chocolatey [master ≡ +0 ~1 -0 !]>
```

PR #182, opened 2017-02-28
(MODULES-4471) Add CI_SPEC_OPTIONS environment variable to modify rspec

glennsarti:
Previously it was not possible to modify the behaviour of RSpec from the rake
tasks.  This commit adds an additional environment variable called
CI_SPEC_OPTIONS which will be appended to the rspec invocation so that additional
settings, such as, output formatting can be passed into the rake tasks.

This commit also adds documentation to the README about CI_SPEC_OPTIONS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,6 +305,31 @@ Testing Parser Functions
 This whole section is superseded by improved support of accessing the scope in
 rspec-puppet.
 
+
+Modify rspec behavior
+================================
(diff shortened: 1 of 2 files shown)


DavidS:
:+1: overall. Please move the CI_NODE_COUNT description from where it is currently to the new section you added. It fits here much better, than under the "Initializing puppet" header.

glennsarti:
@davidS Updated.

PR #186, opened 2017-03-28
Add dependency on parallel_tests

ekohl:
The rake task release_checks uses the parallel_tests so this should be
either a dependency or the task should fall back on regular tests. Given
parallel_tests works well, I decided on the former.

Note that since parallel_tests depends on ruby >= 2.0 this also means
puppetlabs_spec_helper depends on that.


ekohl:
IMHO this is a trivial patch since it's only a metadata change.

PR #188, opened 2017-04-06
[WIP][FM-6131] Implement changelog rake task to autogen CHANGELOG.md

wilson208:
This can be used by specifying the LAST_TAG env var when running the rake task to tell the generator which tag is the latest release tag before the current, pending release. If I was to generate a CHANGELOG.md file for version '2.0.1' I would do the following:
`CHANGELOG_GITHUB_TOKEN={YOUR PERSONAL ACCESS TOKEN} LAST_TAG=2.0.0 bundle exec rake changelog`

It is possible to specify the following env vars also:
GITHUB_PROJECT - Defaults to metadata.json `name` field
GITHUB_USER - Defaults to `puppetlabs`

The above env vars produce the github project url:
https://github.com/[GITHUB_USER]/[GITHUB_PROJECT]

The github_changelog_generator uses octokit under the hood and as such interfaces heavily with the GitHub API. Only 50 requests/hour are permitted against the github API when not using a `Personal Access Token` from GitHub. To specify this, use the `CHANGELOG_GITHUB_TOKEN` env var. One run of the changelog generator will likely exceed 50 API calls so this is a necessary step.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -581,6 +581,39 @@ rescue LoadError
   end
 end
 
+begin
+  require 'github_changelog_generator/task'
+  require 'puppet_blacksmith/rake_tasks'
+
+  GitHubChangelogGenerator::RakeTask.new :changelog do |config|
+    metadata = (Blacksmith::Modulefile.new).metadata
+    header = <<-HERE
+# Change log
+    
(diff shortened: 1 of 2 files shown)


wilson208:
Due to issues in the `github-changelog-generator` gem with the functionality we are looking to use, this relies on fixes made in [this PR](https://github.com/skywinder/github-changelog-generator/pull/453) which is pending merge & release of the `github-changelog-generator` gem.

PR #189, opened 2017-04-17
(MODULES-4688) Release 2.1.2

hunner:


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -371,7 +371,7 @@ def max_thread_limit
     ParallelTests::CLI.new.run(args)
     Rake::Task[:spec_clean].invoke
   rescue LoadError
-    fail 'Add the parallel_tests gem to Gemfile to enable this task'
+    raise 'Add the parallel_tests gem to Gemfile to enable this task'
```
eputnam: what about wrapping the whole task in an `if` statement so it's not even loaded if the gem isn't present? or is a failure better?

hunner: I think I'd rather have failure so that anyone poking through the tasks could find it and figure out how to get it running.

jearls:
Thanks!

PR #190, opened 2017-05-08
(FM-6170) Addition of branch check for build number creation

HelenCampbell:
The build ID on the release pipelines were occasionally causing the builds to break due to having the same or an already existing build number on master job. This PR provides the addition of a simple branch check followed by adding an 'r' to the release branch jobs - ensuring the conflict doesn't happen again.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -500,10 +500,18 @@ task :compute_dev_version do
   end
 
   sha = `git rev-parse HEAD`[0..7]
+  branch = `git rev-parse --abbrev-ref HEAD`
 
   # If we're in a CI environment include our build number
+  # If the branch is a release branch we append an 'r' into the new_version,
+  # this is due to the release branch buildID conflicting with master branch when trying to push to the staging forge.
+  # More info can be found at https://tickets.puppetlabs.com/browse/FM-6170
   if build = ENV['BUILD_NUMBER'] || ENV['TRAVIS_BUILD_NUMBER']
-    new_version = sprintf('%s-%04d-%s', version, build, sha)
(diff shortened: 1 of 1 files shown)


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -500,10 +500,15 @@ def max_thread_limit
   end
 
   sha = `git rev-parse HEAD`[0..7]
+  branch = `git branch`
```
glennsarti: Might be better to use `git rev-parse --abbrev-ref HEAD`

Doesn't require any magic text parsing.

DavidS:
Please use `git rev-parse --symbolic-full-name HEAD` instead of `git branch` to avoid depending on git human readable "porcelain" output.

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -500,10 +500,15 @@ def max_thread_limit
...
 
   # If we're in a CI environment include our build number
   if build = ENV['BUILD_NUMBER'] || ENV['TRAVIS_BUILD_NUMBER']
-    new_version = sprintf('%s-%04d-%s', version, build, sha)
+    if branch.include? "* release"
```
glennsarti: Should we really be special casing release branch? Can we make this generic?

e.g. for master, append nothing. for any other branch append, say the first 7 letters or something?

HelenCampbell: I don't think there's any reason to make this generic? We only ever juggle the two branches on CI. If someone else adds another job they'll need to decide for themselves if it requires the amendment. Though I'll maybe add a comment saying why it's needed so others know what's going on here :)

HelenCampbell:
@glennsarti @DavidS

PR #191, opened 2017-05-31
(maint) Properly escape paths

samuelson:
If the expanded path includes a space the puppet module tool would
interpret it as additional arguments. This change adds quotes around
the paths so that they're interpreted as single arguments.


johnduarte:
@hunner or @DavidS are you the gatekeepers for this repo?
This fixes Jenkins failure seen here:
https://jenkins-imaging.delivery.puppetlabs.net/job/Education%20VMs%20-%20Test/69/console

johnduarte:
@samuelson could you modify your commit message to prepend it with `(maint)`?

i.e.
```
(maint) Properly escape paths
```

PR #192, opened 2017-06-06
(SDK-268) Create directory junctions instead of symlinks on windows

rodjek:
Expands on the windows symlink functionality by falling back to using directory junctions if `Puppet::FileSystem.symlink` fails to create the symlink (or isn't available). Unlike symlinks, NTFS junctions don't require administrator access 🎉 and provide identical behaviour for our purposes.

As `Puppet::FileSystem.symlink` doesn't raise an exception when run as non-administrator, we can just check for the existence of the link target after it (`File.symlink?` conveniently returns `true` for junctions) in order to fall back to using junctions.

The win32-dir gem extends `Dir` with a `create_junction` method that we try to use first. If that gem isn't available then we fall back to shelling out to `mklink` to create the junction.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -240,11 +240,12 @@ task :spec_prep do
   # Ruby only sets File::ALT_SEPARATOR on Windows and Rubys standard library
   # uses this to check for Windows
   is_windows = !!File::ALT_SEPARATOR
-  puppet_symlink_available = false
-  begin
-    require 'puppet'
-    puppet_symlink_available = Puppet::FileSystem.respond_to?(:symlink)
-  rescue
+  if is_windows
+    begin
+      require 'win32/dir'
+    rescue LoadError
+      $stderr.puts "win32-dir gem not installed, falling back to executing mklink directly"
+    end
   end
 
   # git has a race condition creating that directory, that would lead to aborted clone operations
@@ -282,11 +283,16 @@ task :spec_prep do
   repositories.each {|remote, opts| opts[:thread].join }
 
   fixtures("symlinks").each do |source, target|
-    if is_windows
-      fail "Cannot symlink on Windows unless using at least Puppet 3.5" if !puppet_symlink_available
-      Puppet::FileSystem::exist?(target) || Puppet::FileSystem::symlink(source, target)
+    unless File.symlink(target)
+      if is_windows
+        if Dir.respond_to?(:create_junction)
+          Dir.create_junction(target, source)
+        else
+          system("call mklink /J \"#{target.gsub('/', '\\')}\" \"#{source.gsub('/', '\\')}\"")
+        end
+      end
     else
-      File::exists?(target) || FileUtils::ln_sf(source, target)
+      FileUtils::ln_sf(source, target)
     end
   end
 


DavidS:
it looks like the symlink code does error on Windows:

```
PS C:\vagrant\testmodule> bundle exec rake spec_prep
rake aborted!
Puppet::Util::Windows::Error: CreateSymbolicLink(spec/fixtures/modules/C:\vagrant\testmodule, C:\vagrant\testmodule, 1):
  The filename, directory name, or volume label syntax is incorrect.
C:/Users/vagrant/gems/ruby/2.3.0/gems/puppet-4.8.1-x64-mingw32/lib/puppet/util/windows/file.rb:87:in `symlink'
C:/Users/vagrant/gems/ruby/2.3.0/gems/puppet-4.8.1-x64-mingw32/lib/puppet/file_system/windows.rb:44:in `symlink'
C:/Users/vagrant/gems/ruby/2.3.0/gems/puppet-4.8.1-x64-mingw32/lib/puppet/file_system.rb:251:in `symlink'
C:/Users/vagrant/gems/ruby/2.3.0/bundler/gems/puppetlabs_spec_helper-285222104625/lib/puppetlabs_spec_helper/rake_tasks.rb:294:in `block (2 levels) in <top (required)>'
C:/Users/vagrant/gems/ruby/2.3.0/bundler/gems/puppetlabs_spec_helper-285222104625/lib/puppetlabs_spec_helper/rake_tasks.rb:291:in `each'
C:/Users/vagrant/gems/ruby/2.3.0/bundler/gems/puppetlabs_spec_helper-285222104625/lib/puppetlabs_spec_helper/rake_tasks.rb:291:in `block in <top (required)>'
C:/Users/vagrant/gems/ruby/2.3.0/gems/rake-12.0.0/exe/rake:27:in `<top (required)>'
Tasks: TOP => spec_prep
(See full trace by running task with --trace)
PS C:\vagrant\testmodule>
```

this is running this code against a repo with no .fixtures.yml. Maybe that's a 2.1 vs 2.3 distinction, but that won't save us long-term.

rodjek:
Hmm, it doesn't error for me on Ruby 2.3, but I am using Puppet 4.10.1 for testing. Will investigate further (it's an easy enough fix to catch the exception above, but i'm curious why it doesn't get raised on my VM)

rodjek:
Found the cause and it looks like [a valid bug in Puppet](https://github.com/puppetlabs/puppet/commit/54a554ee448097e86fe1e1d0c0896ac57b23dcbb), so time well spent.

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -282,11 +283,16 @@ def max_thread_limit
   repositories.each {|remote, opts| opts[:thread].join }
 
   fixtures("symlinks").each do |source, target|
-    if is_windows
-      fail "Cannot symlink on Windows unless using at least Puppet 3.5" if !puppet_symlink_available
-      Puppet::FileSystem::exist?(target) || Puppet::FileSystem::symlink(source, target)
+    unless File.symlink(target)
```
Iristyle: Did you mean `File.symlink?` here?

This may have been fixed in later Ruby versions, but IIRC, symlink support was pretty broken in Ruby on Windows in earlier versions.

PR #193, opened 2017-06-06
(SDK-168) Replace check:symlinks with platform independent alternative

rodjek:
Shelling out to `find` to check for symlinks won't work well for our Windows friends, so this PR replaces it with a pure Ruby implementation that works on *nix and Windows.

On Linux:
```
asmodean :1: puppet/puppetlabs-ntp git:(master)!» bundle exec rake check:symlinks
Symlink found: spec/fixtures/modules/my_ntp => /home/tsharpe/code/puppet/puppetlabs-ntp/spec/fixtures/my_ntp
Symlink found: spec/fixtures/modules/ntp => /home/tsharpe/code/puppet/puppetlabs-ntp
rake aborted!
Symlink(s) exist within this directory
```

On Windows:
```
C:\Users\tsharpe\code\puppetlabs-ntp>bundle exec rake check:symlinks
Symlink found: spec/fixtures/modules/my_ntp => C:/Users/tsharpe/code/puppetlabs-ntp/spec/fixtures/my_ntp
Symlink found: spec/fixtures/modules/ntp => C:/Users/tsharpe/code/puppetlabs-ntp
rake aborted!
Symlink(s) exist within this directory
```

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -3,6 +3,7 @@ require 'rake'
 require 'rspec/core/rake_task'
 require 'tmpdir'
 require 'yaml'
+require 'pathname'
 
 # optional gems
 begin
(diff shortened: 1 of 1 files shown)


glennsarti:
Should there be tests for this?

Did you try this with both hard links and junctions?
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365006(v=vs.85).aspx

Also here's the symlink detection code in Puppet https://github.com/puppetlabs/puppet/blob/ce5500dd660019b2be6d132b14f70baff4c553fa/lib/puppet/util/windows/file.rb#L213-L220

rodjek:
Yep, junctions are detected fine with the standard `File.symlink?` or `Pathname#symlink?`. I haven't tested hard links as the original find based implementation didn't check for them (would've needed something like `find . -type f -links +1`).

If hard links are something we want to test for, we should probably also change the name of the rake task :)

DavidS:
We specifically check for symlinks, to avoid issues on windows when cloning the repository. tbh, after the recent discoveries around junctions, I believe this to be a git bug, that it uses symlinks, and not junctions, but that's a battle I'm willing to fight.

DavidS:
https://github.com/puppetlabs/puppetlabs_spec_helper/pull/193/files#diff-c7beddb7111031927788e851096dc9deR545 excludes all `.git` directories instead of just the top-level, but I think this is fine.

PR #199, opened 2017-07-14
Support CI options for parallel_spec

austb:


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -372,8 +372,9 @@ def max_thread_limit
   begin
     require 'parallel_tests'
 
-    args = ['-t', 'rspec'].
-      concat(Rake::FileList[pattern].to_a)
+    args = ['-t', 'rspec']
+    args = args.push('--').concat(ENV['CI_SPEC_OPTIONS'].split 'j').push('--') unless ENV['CI_SPEC_OPTIONS'].nil?
```
bmjen: These next two lines don't need the `=` assignment, push and concat should modify the object directly.

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -372,8 +372,9 @@ def max_thread_limit
   begin
     require 'parallel_tests'
 
-    args = ['-t', 'rspec'].
-      concat(Rake::FileList[pattern].to_a)
+    args = ['-t', 'rspec']
+    args.push('--').concat(ENV['CI_SPEC_OPTIONS'].split(' ')).push('--') unless ENV['CI_SPEC_OPTIONS'].nil?
```
bmjen: The guard needs to also protect against an empty string or string of only whitespace. Otherwise you'll just be adding `--` to args without actual args.

PR #206, opened 2017-08-15
Fix 'abort: please specify just one revision' mercurial error

pegasd:
This commit fixes the 'abort: please specify just one revision' when using constructs like the following in .fixtures.yml:

```
my_module:
  repo: "https://hg.mycompany.com/puppet/my_module"
  scm: "hg"
  ref: "0.1.0"
```


DavidS:
Thanks a lot!

PR #207, opened 2017-08-23
Adding the ability to install fixtures from Puppetfile

danzilio:
I frequently use this inside the root of my control repo to test my roles and profiles. Instead of having to list my dependencies in my `Puppetfile` and `.fixtures.yml` I'd like to be able to get my fixtures from my `Puppetfile.` This is a first stab at that.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -66,6 +66,23 @@ def repositories
   @repositories
 end
 
+def puppetfile
+  unless @puppetfile
+    @puppetfile = fixtures('puppetfile')
+    return nil if @puppetfile.empty?
+    require 'r10k/action/puppetfile/install'
+
+    opts = {
+      root: './spec/fixtures',
+      moduledir: './spec/fixtures/modules',
+      puppetfile: @puppetfile
+    }
+
+    @puppetfile = R10K::Action::Puppetfile::Install.new(opts, [])
+  end
+  @puppetfile
+end
+
 # get the array of Beaker set names
 # @return [Array<String>]
 def beaker_node_sets
@@ -127,6 +144,7 @@ def fixtures(category)
 
   result = {}
   if fixtures.include? category and fixtures[category] != nil
+    return fixtures[category] if category == 'puppetfile'
     fixtures[category].each do |fixture, opts|
       if opts.instance_of?(String)
         source = opts
@@ -257,6 +275,8 @@ task :spec_prep do
 
   # git has a race condition creating that directory, that would lead to aborted clone operations
   FileUtils::mkdir_p("spec/fixtures/modules")
+
+  puppetfile.call if puppetfile
 
   repositories.each do |remote, opts|
     scm = 'git'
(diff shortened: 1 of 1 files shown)


DavidS:
Do I understand this correctly, that this adds a `puppetfile:` section to the `.fixtures.yml`, and loads all modules from there using r10k, if it is specified, in addition to all other fixtures?

I think I like it. And the implementation doesn't too bad either.

It still needs a README entry.

danzilio:
@DavidS that's exactly what this does! The `r10k` run happens before the other fixtures get installed, to avoid `r10k` purging the `fixtures/modules` directory.

I'll document this today! Do you think we should add `r10k` as a gem dependency?

DavidS:
thanks!

it might make sense to make the r10k call optional on r10k being installed in the environment so that the dependencies do not explode.

rnelson0:
Could r10k be avoided if it dynamically turned module references into YAML used by fixtures? The existing spec_prep task could probably be modified to use stdin instead of a file.

danzilio:
I hadn't even thought of that, but it's a great idea! I'll take a stab at it!

rnelson0:
Ah, reading the code maybe not YAML but update the results hash directly.

logicminds:
I have a different approach if it is needed/wanted.  
 
https://gitlab.com/nwops/puppet_ci_scripts/blob/master/convert_puppetfile.rb

This converts an existing Puppetfile to a fixtures file. There is no dependencies of R10k as it parses manually without eval. ATM it doesn't support symlinks, paths or forge modules but that can be added. Once the fixtures file exists the spec_helper does the fixtures download.

This feature is badly needed. Any implementation is a step forward.   Does r10k download from forge?  Does r10k install in parallel?  R10k does a great job at caching repos which the fixtures download lacks.

DavidS:
Since my original comment we started shipping r10k with the pdk (via litmus -> bolt -> r10k). I see no reason not to go forward with the originally proposed solution. 

- [ ] rebase this change on top of current master
- [ ] add r10k as a dependency to the gemspec
- [ ] add an entry to the README describing how to use this

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -66,6 +66,23 @@ def repositories
   @repositories
 end
 
+def puppetfile
+  unless @puppetfile
+    @puppetfile = fixtures('puppetfile')
+    return nil if @puppetfile.empty?
+    require 'r10k/action/puppetfile/install'
```
scotje: Either the gemspec needs to be updated to depend directly on `r10k` or this needs to catch `LoadError` exception and handle the case of `r10k` being not installed. (I would prefer the latter.)

I know @DavidS indicated that current versions of PDK have non-explicit dependencies on `r10k` through Litmus and Bolt, but PDK is definitely not the only usage of this gem and I'm also uncomfortable guaranteeing that PDK will always and forever include `r10k` going forward.

logicminds:
Not sure if it is of concern.  Since R10k evals the Puppetfile, any arbitrary ruby code in the Puppetfile would be executed. This may have an unwanted impact, but 80% of the time it should work every time.  This should be stated in the docs.

scotje:
@logicminds Yes, I think it would be good to document that. Speaking of which @danzilio, we should include some documentation for this in the README as well.

daianamezdrea:
Hello @danzilio, are you still working at this PR? Thank you!

pmcmaw:
Due to inactivity I am going to close this PR. 
If you disagree with my decision, please feel free to reopen it and I apologize.

As always, I would like to thank you for taking the time to contribute.

Many Thanks
Paula

PR #211, opened 2017-09-12
Fix release_checks without parallel_tests

sean797:
:release_checks attempts :parallel_spec then :spec if it fails, with https://github.com/puppetlabs/puppetlabs_spec_helper/commit/920d64798ac915c2ffea7434d04e0717ee3251f1 which causes :spec_clean to be invoked twice if parallel_tests is not installed, but crucially it doesn't execute it twice and therefore causes check:symlinks to fail every time when ran via :release_checks without parallel_tests gem

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -389,19 +389,20 @@ end
 desc "Parallel spec tests"
 task :parallel_spec do
   begin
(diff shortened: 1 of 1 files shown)


DavidS:
I've updated this with a bit of cleanup.

@sean797 @ekohl I'd be grateful if one of you could test if that is actually working before we get this merged.

sean797:
Yes, that works for me.

PR #214, opened 2017-10-20
(WIN-6) trim whitespace from test_tiers before parsing

ThoughtCrhyme:


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -61,6 +61,7 @@
     test_tiers = ENV['TEST_TIERS'].split(',')
     raise 'TEST_TIERS env variable must have at least 1 tier specified. low, medium or high (comma separated).' if test_tiers.count == 0
     test_tiers.each do |tier|
+      tier.to_s.strip
```
eputnam: this is pretty nitpicky, but is `to_s` required here?
```
[2] pry(main)> aryy = [ENV['TEST']]
=> ["testing"]
[3] pry(main)> aryy.each do |i|
[3] pry(main)*   puts i.class
[3] pry(main)* end
String
=> ["testing"]
```

PR #216, opened 2017-11-21
bugfix - parallel_spec fails with no files to test

tphoney:


bastelfreak:
Tested this locally, works fine:

```
$ bundle exec rake parallel_spec
(in /home/bastelfreak/modulesync_config/modules/puppet-zabbix)
Could not find semantic_puppet gem, falling back to internal functionality. Version checks may be less robust.
No files for parallel_spec to run against
```

PR #218, opened 2017-11-29
Add `defaults` section to `.fixtures.yml`

jearls:
Add a `defaults` section to the `.fixtures.yml` file.  The `defaults`
section can contain default options for each category of fixture.
Merge those defaults with the "global" defaults used in the `spec_prep`
Rake task, then, for each fixture being installed, merge the fixture's
options into the defaults to get the final set of options for that
fixture.

Use case:  Specifying the same `flags` option for each forge module:

```
defaults:
  forge_modules:
    flags: "--module_repository=https://myforge.example.com/"
fixtures:
  symlinks:
    mymodule: "#{source_dir}"
  forge_modules:
    concat: "puppetlabs/concat"
    stdlib: "puppetlabs/stdlib"
    mysql:
      repo: "puppetlabs/mysql"
      flags: ~ # I need mysql to come from the real Forge
```

Changes:

* `fixtures(category)` updated to use `defaults` section
* `repositories` cleaned up now-obsolete code
* `spec_prep` updated processing of `fixtures(symlinks)` to expect
  a hash with symlink location in `['target']`
* `spec_clean` updated fixtures processing for all categories to
  expect a hash

PR #222, opened 2017-12-09
Fix fixtures handling for missing file or missing `fixtures:` section

jearls:
After submitting PR #218 to add the defaults: section to the .fixtures.yml file, handling of missing .fixtures.yml files or missing fixtures: sections broke.

This PR fixes that in the simplest way - by ensuring that `fixtures` is always a hash.


jearls:
Closing this in favor of #220 which is the same solution

PR #223, opened 2017-12-12
Fix fixtures defaults and add tests

hunner:
The rake tasks methods needed to be moved to a module for spec testing.

No whitespace diff: https://github.com/puppetlabs/puppetlabs_spec_helper/pull/223/files?w=1

For #220 and #221

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -4,6 +4,7 @@ require 'rspec/core/rake_task'
 require 'tmpdir'
 require 'yaml'
 require 'pathname'
+require 'puppetlabs_spec_helper/version'
 
 # optional gems
 begin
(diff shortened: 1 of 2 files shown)


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -75,217 +76,246 @@
...
-  else
-    fixture_defaults = {}
-  end
+  def fixtures(category)
+    if ENV['FIXTURES_YML']
+      fixtures_yaml = ENV['FIXTURES_YML']
+    elsif File.exists?('.fixtures.yml')
+      fixtures_yaml = '.fixtures.yml'
+    elsif File.exists?('.fixtures.yaml')
+      fixtures_yaml = '.fixtures.yaml'
```
bastelfreak: it would be nice if we could move to .yaml, but that will probably never happen.

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -4,6 +4,7 @@
 require 'tmpdir'
 require 'yaml'
 require 'pathname'
+require 'puppetlabs_spec_helper/version'
```
eputnam: is this needed? i don't see the version referenced anywhere

PR #227, opened 2018-02-14
(DO NOT MERGE) Removes the unit test run from the release_checks task

HelenCampbell:
Due to the way we are currently using release checks we have decided to remove the spec test run and keep the checks separate.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -613,11 +613,6 @@ desc "Runs all necessary checks on a module in preparation for a release"
 task :release_checks do
   Rake::Task[:lint].invoke
   Rake::Task[:validate].invoke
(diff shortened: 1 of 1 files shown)


tphoney:
So release checks, no longer runs any unit tests ?
that is how we run unit tests in travis https://github.com/puppetlabs/puppetlabs-apache/blob/master/.travis.yml#L6

HelenCampbell:
Added a do not merge, but module conversion is adding unit test runs as bundle exec rake spec. Release_checks tends to swallow up any error output which makes it quite hard to deal with - I think we should still consider removing the spec tests from this, after we get our modules converted, to ensure we are running both checks and tests seperately. Though willing to be convinced otherwise.  cc/ @pmcmaw for visibility

DavidS:
cc @puppetlabs/pdk

HelenCampbell:
Closing as this is no longer relevant

PR #228, opened 2018-02-15
fixture_rebuild task

eputnam:
closes #109
These are the changes from #109 with a rebase.


eputnam:
@DavidS it doesn't. It looks like @rnelson0 's gem does do what we need with the puppetfile, but I'm not really into adding more dependencies. The `include_metadata` piece shouldn't be much work, though.

DavidS:
Reading through #109 I realised this is not a new idea. See [this comment](https://github.com/puppetlabs/puppetlabs_spec_helper/pull/109#issuecomment-125885255) for more details on how this should be implemented

daianamezdrea:
Hi @eputnam, thank you for your work ! Given the length of time that this has been left to lie fallow and the conversation above, I'll close this PR. Cheers!

PR #229, opened 2018-02-15
(FM-6795) prefer ruby > 2.1.9

eputnam:
This removes some logic and updates some dependencies to drop support for ruby versions less than 2.1.9. This was mainly driven by a sec vulnerability found in an old version of rubocop. In order to bump rubocop, we had to get rid of ruby 1.9.3

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,14 +4,11 @@ source 'https://rubygems.org'
 gemspec
 
 group :development do
-  gem "puppet", ENV['PUPPET_GEM_VERSION'] || ENV['PUPPET_VERSION'] || '~> 4.0'
+  gem "puppet", ENV['PUPPET_GEM_VERSION'] || ENV['PUPPET_VERSION'] || '~> 5.0'
   gem 'rubocop'
   gem 'rubocop-rspec', '~> 1.6' if RUBY_VERSION >= '2.3.0'
 end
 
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -1,22 +1,17 @@
...
-    gem 'rubocop-rspec', '~> 1.6' if RUBY_VERSION >= '2.3.0'
-  end
+  gem 'rubocop'
+  gem 'rubocop-rspec', '~> 1.6' if RUBY_VERSION >= '2.3.0'
 end
 
 # json_pure 2.0.2 added a requirement on ruby >= 2. We pin to json_pure 2.0.1
 # if using ruby 1.x
-gem 'json_pure', '<=2.0.1', :require => false if RUBY_VERSION =~ /^1\./
+gem 'json_pure', '<=2.0.1'
```
bastelfreak: Not sure if this dependency is needed at all. And if it's needed, you can probably stick to a newer version instead of enforcing this old one.

eputnam: good call, i didn't read the comment 🗡

DavidS:
Likely you'll want to exclude the development group from the travis bundle install, to avoid the issues here.

cardil:
You don't need to drop Ruby 1.9 because of Rubocop. You can add newer version of Rubocop only if suitable (ruby >= 2.1) and invoke it only then. See my PR for that: #246

DavidS:
This has been superseded by #246 thanks to @cardil's work.

PR #230, opened 2018-02-16
(FM-6813) fix parsing for test tiers

ThoughtCrhyme:


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -61,16 +61,14 @@
   t.pattern = 'spec/acceptance'
   # TEST_TIERS env variable is a comma separated list of tiers to run. e.g. low, medium, high
   if ENV['TEST_TIERS']
-    tiers = '--tag '
-    test_tiers = ENV['TEST_TIERS'].split(',')
+    test_tiers = test_tiers.split(',')
```
glennsarti: This is splitting on itself.  Shouldn't this still be `test_tiers = ENV['TEST_TIERS'].split(',')` ?

glennsarti:
Nit: Commit message could be better.

DavidS:
What @glennsarti said.

The amount of back and forth over this makes me wonder if there is a way to (unit) test this code on travis.

lib/puppetlabs_spec_helper/rake_tasks.rb
ThoughtCrhyme: Yup, I removed a gsub call from earlier that botched this up. Thanks, Glenn

ThoughtCrhyme:
@DavidS agreed, I'll look into it

PR #234, opened 2018-03-12
Break out beaker and fixture tasks into separate files

DavidS:
This makes them easier to consume on their own without pulling in
everything. This also moves around the tests for those parts, and
adds smoke tests to make sure Rakefiles still load and process
correctly.


DavidS:
See https://github.com/puppetlabs/puppet-resource_api/pull/35 for the reason to have this.

DavidS:
This now depends on #236

DavidS:
Closing this temporarily to rename the source branch.

PR #239, opened 2018-04-09
(PDK-916) Default to mocha if mock_framework isn't set

MikaelSmith:
The mocha framework has been the most common option for module testing.
A recent commit to address issues with mocha 1.5 added a check for which
framework is configured, but not everyone has explicitly configured a
framework. Check explicitly for rspec, and fall-back to mocha otherwise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -463,5 +463,52 @@ The following links may give you some insight into why...
 
 [Microsoft TechNet](https://technet.microsoft.com/en-us/library/cc754077.aspx)
 
+mock_with
+=========
+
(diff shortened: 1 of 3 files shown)


DavidS:
@MikaelSmith amended the check to actually work, and expanded on your idea of only acting if no mock framework has been chosen. See also https://tickets.puppetlabs.com/browse/PDK-916

MikaelSmith:
I'm fine with this being taken over. Need anything more from me?

DavidS:
Thanks for you input, and patience, @MikaelSmith !

PR #241, opened 2018-05-08
(PDK-636) Always remove symlink fixtures. Only remove downloaded fixtures if tests pass.

rodjek:
Restores most of the original behaviour of running `spec`, so that downloaded fixtures are only cleaned up if the tests pass. Fixtures created from symlinks (which cause problems when kept around) are still always removed at the end of each test run regardless of pass or fail.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -60,8 +60,9 @@ task :spec do |t, args|
   begin
     Rake::Task[:spec_prep].invoke
     Rake::Task[:spec_standalone].invoke(*args.extras)
-  ensure
     Rake::Task[:spec_clean].invoke
+  ensure
+    Rake::Task[:spec_clean_symlinks].invoke
   end
 end
 
@@ -78,8 +79,9 @@ task :parallel_spec do
 
       Rake::Task[:spec_prep].invoke
       ParallelTests::CLI.new.run(args)
-    ensure
       Rake::Task[:spec_clean].invoke
+    ensure
+      Rake::Task[:spec_clean_symlinks].invoke
     end
   end
 end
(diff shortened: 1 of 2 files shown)


farkasmate:
For me it sounds like it's not fulfilling the second acceptance criteria:
`on a module with fixtures, running pdk test unit a second time has much lower setup time`

`pdk test unit` using PDK 1.4.1 takes about 10 seconds with an empty class and a generated test to finish.
If I'm adding `puppetlabs-stdlib` and `puppetlabs-concat` as dependencies the same run takes about 24 seconds to finish.

In my reading the ticket is about to make the second execution time close to the execution time of a test without dependencies - 10 seconds in the example above.

DavidS:
I've updated PDK-636 to reflect the current status of development of this improvement accurately.

rodjek:
@farkasmate are those modules being pulled in from git or the forge?

farkasmate:
Forge.

```
---
fixtures:
  forge_modules:
    concat:
      repo: 'puppetlabs-concat'        
      ref: '4.2.0'
    stdlib:
      repo: 'puppetlabs-stdlib'        
      ref: '4.24.0'
```

alexharv074:
This change is going to change the behaviour of the spec_clean task, which is going to confuse people I think, considering it's had the same behaviour for many years.

farkasmate:
@alexharv074 I agree, it might get confusing.

I prefer the current behaviour in CI, but for local runs I'd like to keep the cached fixtures to speed up additional runs.

Ideally I'd trigger the new behaviour by specifying an option on the command line or setting a value in a config file under `~/.pdk/`.

rodjek:
Yeah, I think it makes sense to see these as separate changes. This change restores the behaviour of the `spec_clean` task to what people are used to. We can add a flag to `pdk test unit` to not clean up the downloaded fixtures (even on a successful run) to handle the quicker local interative workflow side of things.

PR #245, opened 2018-06-15
(maint) adding ruby code coverage setup and rake task

tphoney:


tphoney:
@DavidS  what do you think ?

DavidS:
Needs to be rebased on top of #246

PR #246, opened 2018-06-15
Stable development structure for this gem

cardil:
* Adding puppet 5 to travis tests.
* Adding rubocop where it's possible.
* Using `Gemfile.lock` to hold gems to their resolved state.
* Adding all existing rubocop violations to TODO file.
* Randomizing test order to detect unit test dependent issues.
* Adding code coverage reporting.


.rubocop.yml
```diff
@@ -1,504 +1,138 @@
...
-Style/AndOr:
-  Enabled: True
-
-Style/RedundantSelf:
-  Enabled: True
-
-# Method length is not necessarily an indicator of code quality
-Metrics/MethodLength:
+AllCops:
+  TargetRubyVersion: 2.1.9
```
cardil: Why have you set target compatibility to ruby 2.1 if you still support Ruby 1.9. I think support for Ruby 1.9 should still be there.

Ruby 1.9 is needed for Puppet 3. This gem is used by virtually all Puppet modules out there and there is many modules that use Puppet 3, even in 2018.

I think Puppet 3 and Ruby 1.9 should stay.

BTW. You have left Ruby 1.9 and Puppet 3 in `.travis-ci.yml` build file.

cardil:
Why have you set target compatibility for Rubocop to ruby 2.1 if you still support Ruby 1.9. I think support for Ruby 1.9 should still be there.

Ruby 1.9 is needed for Puppet 3. This gem is used by virtually all Puppet modules out there and there is many modules that use Puppet 3, even in 2018.

I think Puppet 3 and Ruby 1.9 should stay.

BTW. You have left Ruby 1.9 and Puppet 3 in `.travis-ci.yml` build file.

cardil:
I have reverted Rubocop code compatibility to Ruby 1.9 in commit 3f0385d. @DavidS if you think it's bad idea I will remove this commit from PR.

DavidS:
Nah, it's fine.

PR #252, opened 2018-06-27
update README

b4ldr:
small documentation updated, i am curios, why limit functionality to just three tiers .

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,8 +346,7 @@ remembering to duplicate any existing environment variables:
       - FUTURE_PARSER=yes CI_NODE_TOTAL=2 CI_NODE_INDEX=2
 
 #### Running tests tagged with test tiers
-To run tests tagged with risk levels set the ``TEST_TIERS`` environment variable to a comma-separated list of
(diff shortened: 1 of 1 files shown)


DavidS:
cc @ThoughtCrhyme for the question.

ThoughtCrhyme:
@b4ldr Hi John, we use three tiers because our internal testing happens in three cadences: per PR, nightly, and weekly corresponding to "high", "medium" and "low" tiers.

b4ldr:
@ThoughtCrhyme it would be useful to be able to use arbitrary tiers here, would you be open to removing the constraint on these keywords or making it possible to configure the [checking array](https://github.com/puppetlabs/puppetlabs_spec_helper/blob/master/lib/puppetlabs_spec_helper/tasks/beaker.rb#L70).  I would be happy to create a PR if your open to it

ThoughtCrhyme:
@b4ldr we'd be happy to review any PR's that make it easier to run the acceptance tests at arbitrary test tiers.

PR #253, opened 2018-06-28
add support to override the allowed test tiers

b4ldr:
This PR adds support for a new environment variable `TEST_TIERS_ALLOWED` to allow users to override the default list of allowed test teirs

### DIFF
--- a/lib/puppetlabs_spec_helper/tasks/beaker.rb
+++ b/lib/puppetlabs_spec_helper/tasks/beaker.rb
@@ -64,10 +64,11 @@ class SetupBeaker
     # TEST_TIERS env variable is a comma separated list of tiers to run. e.g. low, medium, high
     if ENV['TEST_TIERS']
       test_tiers = ENV['TEST_TIERS'].split(',')
-      raise 'TEST_TIERS env variable must have at least 1 tier specified. low, medium or high (comma separated).' if test_tiers.count == 0
+      test_tiers_allowed = ENV.fetch('TEST_TIERS_ALLOWED', 'low,medium,high').split(',')
+      raise 'TEST_TIERS env variable must have at least 1 tier specified. Either low, medium or high or one of the tiers listed in TEST_TIERS_ALLOWED (comma separated).' if test_tiers.count == 0
       test_tiers.each do |tier|
         tier_to_add = tier.strip.downcase
-        raise "#{tier_to_add} not a valid test tier." unless %w[low medium high].include?(tier_to_add)
+        raise "#{tier_to_add} not a valid test tier." unless test_tiers_allowed.include?(tier_to_add)
         tiers = "--tag tier_#{tier_to_add}"
         t.rspec_opts.push(tiers)
       end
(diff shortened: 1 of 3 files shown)


spec/unit/puppetlabs_spec_helper/tasks/beaker_spec.rb
```diff
@@ -25,5 +25,21 @@
       allow(ENV).to receive(:[]).and_return('"high", "medium", "low"')
       expect { described_class.setup_beaker(task) }.to raise_error(RuntimeError, %r{not a valid test tier})
     end
+    it 'errors when tiers are not in the allowe list' do
+      allow(ENV).to receive(:[]).and_return('foobar')
+      expect { described_class.setup_beaker(task) }.to raise_error(RuntimeError, %r{not a valid test tier})
+    end
+    it 'Override TEST_TIERS_ALLOWED' do
+      ENV['TEST_TIERS_ALLOWED'] = 'dev,rnd'
```
DavidS: Please set up mocking like in the next line, instead of altering the global environment.

Background: using the `allow(ENV).to receive(:[]).with('TEST_TIERS').and_return('dev')` syntax means that everything is cleaned up in all cases. Modifying global state means that when errors happen, cleanup is skipped, leading to flaky tests.

lib/puppetlabs_spec_helper/tasks/beaker.rb
```diff
@@ -64,10 +64,11 @@ def self.setup_beaker(t)
     # TEST_TIERS env variable is a comma separated list of tiers to run. e.g. low, medium, high
     if ENV['TEST_TIERS']
       test_tiers = ENV['TEST_TIERS'].split(',')
-      raise 'TEST_TIERS env variable must have at least 1 tier specified. low, medium or high (comma separated).' if test_tiers.count == 0
+      test_tiers_allowed = ENV.fetch('TEST_TIERS_ALLOWED', 'low,medium,high').split(',')
```
DavidS: :+1: !

b4ldr:
@DavidS My rspec foo is not that good and i was unable to get it to work using the following 

```ruby
    it 'Override TEST_TIERS_ALLOWED' do
      allow(ENV).to receive(:[]).with('TEST_TIERS_ALLOWED').and_return('dev,rnd')
      allow(ENV).to receive(:[]).with('TEST_TIERS').and_return('dev')
      expect(described_class.setup_beaker(task).rspec_opts.to_s).to match(%r{--tag tier_dev})
    end
```

Im not entirely sure why `allow(ENV).to receive(:[]).and_return('foobar')` works without using the `.with` modifier.  if you are able to give some pointers i would be happy to make the changes

Thanks John

DavidS:
@b4ldr I'll have :eyes: on that tomorrow.

DavidS:
ah, actually, that's easy. because your code is using `ENV.fetch()`, not `ENV[]`. If you want to play a bit more with it feel free, otherwise I can fix it up easily tomorrow.

b4ldr:
@DavidS thanks that was an easy one :), let me know if you want any other changes

DavidS:
Thank you for your time and work!

PR #254, opened 2018-08-24
Update documentation for older Puppet versions

op-ct:
This patch documents which environment variables only apply to Puppet 3.x, and updates broken documentation links with the best available version.


tphoney:
thanks for the change @op-ct !!!

PR #255, opened 2018-08-29
(feat) add puppet lint fix task

tphoney:


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -144,6 +144,28 @@
...
+    documentation
+    single_quote_string_with_variables
+  ]
+  config.ignore_paths = [
+    'bundle/**/*.pp',
+    'pkg/**/*.pp',
+    'spec/**/*.pp',
+    'tests/**/*.pp',
+    'types/**/*.pp',
+    'vendor/**/*.pp',
```
bastelfreak: on Vox Pupuli, we also ignore `.vendor`. Maybe you want to add that as well:
https://github.com/voxpupuli/modulesync_config/blob/5a7d028a95278d11c2b02616d300fda002fefa0d/moduleroot/Rakefile.erb#L22

tphoney:
thanks @HelenCampbell @bastelfreak , what do you think ??

PR #257, opened 2018-09-18
Fix example conversion from mocha to rspec mocks.

vStone:
rspec mocks uses `and_return`, not `and_returns`


DavidS:
Thanks @vStone !

PR #258, opened 2018-09-21
(MODULES-7856) Allow optional repositories based on puppet version

joshcooper:
Some types/providers maybe missing when testing against puppet6 as a
gem. Allow an optional repository to be specified which will only be
installed if the current puppet version matches the semantic version
constraint. For example, stdlib will always be installed, but
selinux_core is only installed when testing on puppet6 or above:

    repositories:
      stdlib: https://github.com/puppetlabs/puppetlabs-stdlib.git
      selinux_core:
        repo: https://github.com/puppetlabs/puppetlabs-selinux_core.git
        puppet_version: ">= 6.0.0"

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ When specifying the repo source of the fixture you have a few options as to whic
    ```yaml
    flags: --verbose
    ```
- * puppet_version - versions of puppet for which the fixture should be installed. Ruby version constraints are supported. Only supported on puppet 4.0 and up.
+ * puppet_version - versions of puppet for which the fixture should be installed. Ruby version constraints are supported. Only works when the `semantic_puppet` gem is available (shipped with puppet 4.0 and up, by default).
 
    ```yaml
    puppet_version: '>= 6.0.0'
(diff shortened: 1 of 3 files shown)


joshcooper:
Looks like it's failing on puppet 3.0 because that version doesn't depend on or vendor semantic_puppet. I'll take a look in a bit.

bastelfreak:
can we please finally drop puppet 3? it is EOL since almost 2 years. is it even on the test mattix on purpose?

joshcooper:
I worked around the puppet 3 issue for now (so the `puppet_version` attribute is ignored). But I am in favor of dropping puppet 3 from the test matrix.

joshcooper:
👍 to dropping puppet 3.x from the travis matrix.

@DavidS does this need a PDK ticket? Should I target an earlier branch?

davidmalloncares:
Hey @joshcooper, would you be able to get a look at this one today around David Schmitts review comments? This change is currently causing issues for Modules teams and Vox for Puppet 6 so would be cool if we can get unblocked asap

joshcooper:
@DavidS this should be good now

* Added a JIRA ticket
* Updated docs about `puppet_version` only working on Puppet 4 and up
* Reworked the code to test for the `SemanticPuppet` being defined, instead of relying on puppet version heuristic
* Added tests

PR #260, opened 2018-10-05
(PDK-1100) Use PDK to build module packages

rodjek:
Attempt to load the PDK gem first in order to use PDK as a library. If
that fails, fall back to executing `pdk build`.

*Note* This depends on PDK 1.7.1 being released and so shouldn't be
merged until then.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -103,14 +103,10 @@ end
 
 desc 'Build puppet module package'
 task :build do
-  begin
-    require 'pdk/module/build'
+  require 'pdk/module/build'
 
-    path = PDK::Module::Build.invoke(:force => true, :'target-dir' => File.join(Dir.pwd, 'pkg'))
-    puts "Module built: #{path}"
-  rescue LoadError
-    system('pdk build --force')
-  end
+  path = PDK::Module::Build.invoke(:force => true, :'target-dir' => File.join(Dir.pwd, 'pkg'))
+  puts "Module built: #{path}"
 end
 
 desc 'Clean a built module package'
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -103,13 +103,13 @@ end
 
 desc 'Build puppet module package'
 task :build do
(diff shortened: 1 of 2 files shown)


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -103,13 +103,13 @@
...
 desc 'Build puppet module package'
 task :build do
-  # This will be deprecated once puppet-module is a face.
   begin
-    Gem::Specification.find_by_name('puppet-module')
-  rescue Gem::LoadError, NoMethodError
-    require 'puppet/face'
-    pmod = Puppet::Face['module', :current]
-    pmod.build('./')
+    require 'pdk/module/build'
```
bmjen: Concerned about having this adding pdk gem as a soft dependency. If the default behavior is to use pdk's build feature as a library method, then perhaps we should add pdk as a runtime dependency.

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -103,13 +103,13 @@
...
+
+    path = PDK::Module::Build.invoke(:force => true, :'target-dir' => File.join(Dir.pwd, 'pkg'))
+    puts "Module built: #{path}"
+  rescue LoadError
+    system('pdk build --force')
```
bmjen: Should we also check if `Puppet::Face['module', :current]` is nil or returns an error? And potentially default to the old way if the puppet < 6 is being loaded?

lib/puppetlabs_spec_helper/rake_tasks.rb
scotje: I'm OK with the soft dependency for accessing it as a library but I think there should be a more explicit error message at the end if `pdk` is not available on their PATH, or fall back to loading the old face which I think is what Bryan is suggesting below?

rodjek:
OK, I've gone back to a soft dependency for now until we can resolve the `json_pure` dependency in PDK that conflicts with Puppet 4.x.

As the logic currently stands:
 * If Puppet < 6.0.0, build with PMT as before
 * If Puppet > 6.0.0
   * If the PDK gem is available, use PDK as a library and call `PDK::Module::Build.invoke`
   * If the PDK gem is not available:
     * If PDK is installed, execute `pdk build --force`
     * If PDK is not installed, print an error asking the user to install PDK.

PR #261, opened 2018-10-08
Added tasks to rspec pattern.

dylanratcliffe:
While I have yet to see anyone testing ruby tasks, I see no reason why we should prevent that.


dylanratcliffe:
Also changed to alphabetical order

PR #266, opened 2018-12-06
(PDK-997) Remove Dir.chdir call from check:test_file task

rodjek:
This `chdir` breaks the isolation of the rake task and affects the
behaviour of other tasks that are invoked in it in the same `rake`
execution.

Rather than modify the task so that it changes back to the parent
directory after the task has finished, we can just remove the `chdir`
entirely and just change the glob to look for files in the directory.

This PR also adds some basic unit tests for the `check:test_file` task.


nkanderson:
@rodjek This seems like a sensible change to me. It's admittedly my first interaction with this codebase though, so feel free to pull in another reviewer if you'd like to get any additional eyes on it.

PR #270, opened 2019-01-10
(MAINT) Add Plans Path Exclusion

RandomNoun7:
The Puppet parser cannot yet handle bolt plan syntax. Without this
exlusion adding plans to a module causes CI to fail in Appveyor because
the default lint checks will fail.


RandomNoun7:
This change is in coordination with a PR to the PDK templates, to ensure that modules that have plans don't need as many hand crafted changes to the template controlled files in the module after every PDK Update. https://github.com/puppetlabs/pdk-templates/pull/176/files

jpogran:
This is pending #271

PR #271, opened 2019-01-10
(maint) - Resolving bundler ruby version failure, updating tests to include puppet 6

pmcmaw:
Removing the pin from bundler as the older version of bundler does not support newer version of ruby. Also updating testing to include puppet 6.

This PR is required to allow #270 to progress.


david22swan:
Look's good to me, but you should probably get the opinion of someone who has worked more with the module.
+1

PR #277, opened 2019-01-23
Update documentation with PR#218

Felixoid:
It was really tricky to find a proper way to define global forge url to local testing. This PR adjust the documentation to #218 and https://tickets.puppetlabs.com/browse/MODULES-6226


DavidS:
Thanks a lot for your contribution!

PR #278, opened 2019-02-14
(maint) migrate the changelog task from pdk-templates

DavidS:
This commit adds the github_changelog_generator task from pdk-templates,
including some cleanups by putting everything into a function.

To access template parameters required for the PDK, everything is wrapped
into the `create_gch_task` which also neatly contains the helper functions.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -361,3 +361,79 @@ if File.exist? locales_dir
     puts 'No gettext-setup gem found, skipping GettextSetup config initialization' if Rake.verbose == true
   end
 end
+
+def create_gch_task(changelog_user = nil, changelog_project = nil, changelog_since_tag = nil, changelog_tag_pattern = 'v%s')
+  if Bundler.rubygems.find_name('github_changelog_generator').any?
+    # needed a place to hide these methods
+    # rubocop:disable Lint/NestedMethodDefinition
+    def changelog_user_from_metadata
+      result = JSON.parse(File.read('metadata.json'))['author']
+      raise 'unable to find the changelog_user in .sync.yml, or the author in metadata.json' if result.nil?
+      puts "GitHubChangelogGenerator user:#{result}"
+      result
+    end
+
+    def changelog_project_from_metadata
+      result = JSON.parse(File.read('metadata.json'))['name']
+      raise 'unable to find the changelog_project in .sync.yml or the name in metadata.json' if result.nil?
+      puts "GitHubChangelogGenerator project:#{result}"
+      result
+    end
+
+    def changelog_future_release
+      return unless Rake.application.top_level_tasks.include? 'changelog'
+      result = JSON.parse(File.read('metadata.json'))['version']
+      raise 'unable to find the future_release (version) in metadata.json' if result.nil?
+      puts "GitHubChangelogGenerator future_release:#{result}"
+      result
+    end
+    # rubocop:enable Lint/NestedMethodDefinition
+
+    GitHubChangelogGenerator::RakeTask.new :changelog do |config|
+      if ENV['CHANGELOG_GITHUB_TOKEN'].nil?
+        raise "Set CHANGELOG_GITHUB_TOKEN environment variable eg 'export CHANGELOG_GITHUB_TOKEN=valid_token_here'"
+      end
+      config.user = changelog_user || changelog_user_from_metadata
+      config.project = changelog_project || changelog_project_from_metadata
+      config.since_tag = changelog_since_tag if changelog_since_tag
+      config.future_release = changelog_tag_pattern % changelog_future_release.to_s
+      config.exclude_labels = ['maintenance']
+      config.header = "# Change log\n\nAll notable changes to this project will be documented in this file. " \
+                      'The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) and this project adheres ' \
+                      'to [Semantic Versioning](https://semver.org).'
+      config.add_pr_wo_labels = true
+      config.issues = false
+      config.merge_prefix = '### UNCATEGORIZED PRS; GO LABEL THEM'
+      config.configure_sections = {
+        'Changed' => {
+          'prefix' => '### Changed',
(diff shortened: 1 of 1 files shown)


DavidS:
To use this, all the changelog cruft in https://github.com/puppetlabs/pdk-templates/blob/7281db56ead0feeacbc70e5d5dfb3bb885c554ba/moduleroot/Rakefile.erb can be replaced by
```
create_gch_task(<%= changelog_user.inspect %>, <%= changelog_project.inspect %>, <%= changelog_since_tag.inspect %>)
```

Expect a PR for that soon[tm].

ekohl:
In the Foreman modulesync we use it slightly different: https://github.com/theforeman/foreman-installer-modulesync/blob/ccc8b5feeb4b66e6e40d7501822d95fb00e414e8/moduleroot/Rakefile.erb#L18-L41. The excludes label is hardcoded now. How about another approach to this: read additional fields from metadata.json. Not sure if the spec allows this, but a generic mapping there would be much more flexible than templating Ruby code. Alternatively you could actually read .sync.yml but that'd be a bit of a hack.

DavidS:
@ekohl This code here does fall back to metadata.json values if no other values are provided. In pdk-templates code generation is preferred over reading from sync.yml, because the latter would exclude people from forking pdk-templates and modifying the config_defaults there.

ekohl:
@DavidS I meant that we could add something in metadata.json:
```json
{
  "puppetlabs_spec_helper": {
    "changelog": {
      "user": "myuser",
      "project": "myproject",
      "exclude_labels": ["ignore", "skip-changelog"]
    }
  }
}
```

This probably means expanding metadata.json, but the big benefit is that it hides much more Ruby code from the Puppet developers. Generating boiler plate is something we should really avoid when we can.

DavidS:
Ah, now I understand. Long-term this'll migrate to https://github.com/puppetlabs/pdk-planning/blob/master/RFCs/0002-add-pdk-config.md; for now, it removes a big codeblock from the module's Rakefile and replaces it with a single line without requiring a migration process.

ekohl:
:+1: I agree, this is short term a very nice solution. Only generating a single line is already a huge improvement in the right direction.

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -361,3 +361,78 @@
     puts 'No gettext-setup gem found, skipping GettextSetup config initialization' if Rake.verbose == true
   end
 end
+
+def create_gch_task(changelog_user, changelog_project, changelog_since_tag, changelog_tag_pattern = 'v%s')
```
ekohl: Could you add `= nil` for first 3 arguments? The code handles it and it'd benefit users who don't want to override anyway.

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -361,3 +361,78 @@
...
+      if Rake.application.top_level_tasks.include?('changelog') && ENV['CHANGELOG_GITHUB_TOKEN'].nil?
+        raise "Set CHANGELOG_GITHUB_TOKEN environment variable eg 'export CHANGELOG_GITHUB_TOKEN=valid_token_here'"
+      end
+      config.user = changelog_user || changelog_user_from_metadata
+      config.project = changelog_project || changelog_project_from_metadata
+      config.since_tag = changelog_since_tag if changelog_since_tag
+      config.future_release = changelog_tag_pattern % changelog_future_release.to_s
+      config.exclude_labels = ['maintenance']
+      config.header = "# Change log\n\nAll notable changes to this project will be documented in this file. " \
+                      'The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres ' \
```
ekohl: Could you use https here?

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -361,3 +361,78 @@
...
+    def changelog_future_release
+      return unless Rake.application.top_level_tasks.include? 'changelog'
+      result = JSON.parse(File.read('metadata.json'))['version']
+      raise 'unable to find the future_release (version) in metadata.json' if result.nil?
+      puts "GitHubChangelogGenerator future_release:#{result}"
+      result
+    end
+
+    GitHubChangelogGenerator::RakeTask.new :changelog do |config|
+      if Rake.application.top_level_tasks.include?('changelog') && ENV['CHANGELOG_GITHUB_TOKEN'].nil?
```
ekohl: Why is the `Rake.application.top_level_tasks.include?('changelog')` part needed here? Doesn't the line above it guarantee that already?

lib/puppetlabs_spec_helper/rake_tasks.rb
DavidS: :+1:

lib/puppetlabs_spec_helper/rake_tasks.rb
DavidS: yes :+1: also in the semver link in the next line.

lib/puppetlabs_spec_helper/rake_tasks.rb
DavidS: good spot. removed.

DavidS:
@puppetlabs/pdk resolved @ekohl 's comments

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -361,3 +361,79 @@
...
+          'prefix' => '### Fixed',
+          'labels' => ['bugfix'],
+        },
+      }
+    end
+  else
+    desc 'Generate a Changelog from GitHub'
+    task :changelog do
+      raise <<EOM
+The changelog tasks depends on unreleased features of the github_changelog_generator gem.
```
ekohl: Recently there was a release after a long time. I haven't tried it yet, but in general that'd be preferred I think.

DavidS: yes, we have our eyes on that. it's a separate task though.

DavidS: (and a lot easier to fix after this been merged and deployed)

PR #274, opened 2019-01-11
Load puppet-strings tasks when available

ekohl:
This provides the string:generate task which can be useful for module developers.

PR #279, opened 2019-02-15
(maint) load rake tasks from optional libraries

DavidS:
This removes the need for the same loading code in pdk-templates at
https://github.com/puppetlabs/pdk-templates/blob/7281db56ead0feeacbc70e5d5dfb3bb885c554ba/moduleroot/Rakefile.erb#L13-L16

Note that `puppet-syntax/tasks/puppet-syntax` was already loaded in this
file further down.


DavidS:
This supercedes #274

PR #281, opened 2019-03-07
(feat) dont load the beaker if litmus is there

tphoney:


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -4,11 +4,13 @@
 require 'tmpdir'
 require 'pathname'
 require 'puppetlabs_spec_helper/version'
-require 'puppetlabs_spec_helper/tasks/beaker'
 require 'puppetlabs_spec_helper/tasks/fixtures'
 require 'puppetlabs_spec_helper/tasks/check_symlinks'
 require 'English'
 
+# dont load beaker if litmus is present
+require 'puppetlabs_spec_helper/tasks/beaker' unless Bundler.rubygems.find_name('puppet_litmus').any?
```
ekohl: This broke: `NameError: uninitialized constant Bundler`

[Rakefile](https://github.com/theforeman/foreman-infra/blob/06ad0cb12d5d4b9d239669b48d20168c984b14f8/puppet/Rakefile) and [Gemfile](https://github.com/theforeman/foreman-infra/blob/06ad0cb12d5d4b9d239669b48d20168c984b14f8/puppet/Gemfile) for reference.

pmcmaw: Thanks for letting us know @ekohl 
We will work on a fix and have it released as soon as we possibly can.

ekohl: Thanks! Perhaps it should check if beaker is actually present? In many cases people just exclude beaker unless they're running acceptance tests in their Gemfiles.

pmcmaw: This issue is being tracked here: https://tickets.puppetlabs.com/browse/MODULES-8778
:-)

PR #283, opened 2019-03-22
Remove codecov from coverage generator

Ultimation:
Ruby default practices should not include third party tools and internet uploads for a simple coverage report. That should be additional functionality.

### DIFF
--- a/lib/puppetlabs_spec_helper/module_spec_helper.rb
+++ b/lib/puppetlabs_spec_helper/module_spec_helper.rb
@@ -22,12 +22,10 @@ if ENV['SIMPLECOV'] == 'yes'
   begin
     require 'simplecov'
     require 'simplecov-console'
-    require 'codecov'
 
     SimpleCov.formatters = [
       SimpleCov::Formatter::HTMLFormatter,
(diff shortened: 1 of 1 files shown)


lib/puppetlabs_spec_helper/module_spec_helper.rb
```diff
@@ -22,12 +22,10 @@ def verify_contents(subject, title, expected_lines)
...
     require 'simplecov-console'
-    require 'codecov'
 
     SimpleCov.formatters = [
       SimpleCov::Formatter::HTMLFormatter,
       SimpleCov::Formatter::Console,
-      SimpleCov::Formatter::Codecov,
     ]
     SimpleCov.start do
       track_files 'lib/**/*.rb'
```
Ultimation: Wait, how did this ever get coverage for puppet files?

DavidS: Not. This is only for ruby code.

daianamezdrea:
Hi @eputnam, thank you for your work ! Given the length of time that this has been left to lie fallow and the conversation above, I'll close this PR. Cheers!

PR #284, opened 2019-03-26
Download forge modules in parallel

logicminds:
This improves the forge module download speed by at least 400% when multiple modules are specified in fixtures.yml. Git repositories were previously already using parallel downloads so this only fixes modules.

This also refactors the fixtures code by adding almost everything into a method and placed inside the fixtures module spec module.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,11 +7,12 @@ gemspec
 group :development do
   gem 'codecov'
   gem 'github_changelog_generator' if Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new('2.2.0')
+  gem 'pry'
   gem 'puppet', ENV['PUPPET_GEM_VERSION'] || ENV['PUPPET_VERSION'] || '~> 4.0'
   gem 'simplecov', '~> 0'
   gem 'simplecov-console'
   if Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new('2.1.0')
-    gem 'rubocop', '< 0.50'
+    gem 'rubocop', '= 0.49'
     gem 'rubocop-rspec', '~> 1'
   end
 end
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Host github.com
 
 ```
 
-Note: parallel downloads is only available for repositories and not forge modules.
+Note: parallel downloads are available for repositories and forge modules.
 
 ### Parallel tests
 It is also possible to use the `parallel_tests` Gem via the `:parallel_spec` Rake task to run rspec commands in parallel on groups of spec files.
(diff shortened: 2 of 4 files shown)


logicminds:
Looks like my editor changed a bunch of markdown stuff too with the vscode markdown plugin.

logicminds:
Not sure what is wrong the the CLA bot but I already signed so I can't sign twice.

Also rubocop is failing without my changes.

Gemfile
```diff
@@ -7,11 +7,12 @@ gemspec
 group :development do
   gem 'codecov'
   gem 'github_changelog_generator' if Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new('2.2.0')
+  gem 'pry'
   gem 'puppet', ENV['PUPPET_GEM_VERSION'] || ENV['PUPPET_VERSION'] || '~> 4.0'
   gem 'simplecov', '~> 0'
   gem 'simplecov-console'
   if Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new('2.1.0')
-    gem 'rubocop', '< 0.50'
+    gem 'rubocop', '= 0.49'
```
jorhett: hard-coding a very specific value risks conflict with other requirements. Why is this specific version necessary?

README.md
```diff
@@ -1,26 +1,23 @@
-Puppet Labs Spec Helper
-=======================
+# Puppet Labs Spec Helper
```
jorhett: I generally feel (as a non-owner of the repo) that non-change diffs should be removed from a PR so as to focus on the changes. A separate PR with the README improvements would make review easier.

lib/puppetlabs_spec_helper/tasks/fixtures.rb
```diff
@@ -249,77 +272,46 @@ def current_thread_count(items)
...
-      @max_thread_limit = if ENV['MAX_FIXTURE_THREAD_COUNT'].to_i > 0
-                            ENV['MAX_FIXTURE_THREAD_COUNT'].to_i
-                          else
-                            10 # the default
-                          end
-    end
-    @max_thread_limit
+    # the default thread count is 10 but can be
+    # raised by using environment variable MAX_FIXTURE_THREAD_COUNT
+    @max_thread_limit ||= ENV['MAX_FIXTURE_THREAD_COUNT'].nil? ? 10 : ENV['MAX_FIXTURE_THREAD_COUNT'].to_i
```
jorhett: This reduces readability. I would suggest implementing the revised code in the more readable format used previously.

Gemfile
logicminds: Without the hard requirement there were dependency issues and rubocop warnings I could not fix without pinning to this version as my rubocop was much older.  It is assumed a developer has the latest rubocop but if 0.41 is used, there are tons of warnings.  We can't use 0.50 due to requirement of newer ruby.

README.md
logicminds: Removed automated conversion as suggested.

baurmatt:
This look really promising! :) What needs to be done to get this merged?

It looks like the failing test is due to a Travis error and unrelated to the code changes.

logicminds:
Probably just a rebase, and a slight code change to accommodate older puppet.  I'll fix this and ping the maintainers.

PR #285, opened 2019-03-26
(MODULES-8778) - Only use Bundler constant if it exists

pmcmaw:
This PR fixes an issue seen by @ekohl where the following error was being thrown: 
`NameError: uninitialized constant Bundler`

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -9,7 +9,7 @@ require 'puppetlabs_spec_helper/tasks/check_symlinks'
 require 'English'
 
 # dont load beaker if litmus is present
-require 'puppetlabs_spec_helper/tasks/beaker' unless Bundler.rubygems.find_name('puppet_litmus').any?
+require 'puppetlabs_spec_helper/tasks/beaker' unless Object.const_defined?(:Bundler) && Bundler.rubygems.find_name('puppet_litmus').any?
 
 # optional gems
 begin


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -9,7 +9,11 @@
 require 'English'
 
 # dont load beaker if litmus is present
-require 'puppetlabs_spec_helper/tasks/beaker' unless Bundler.rubygems.find_name('puppet_litmus').any?
+if Object.const_defined?(:Bundler)
+  require 'puppetlabs_spec_helper/tasks/beaker' unless Bundler.rubygems.find_name('puppet_litmus').any?
+else
+  require 'puppetlabs_spec_helper/tasks/beaker'
```
tphoney: `require 'puppetlabs_spec_helper/tasks/beaker' unless  Object.const_defined?(:Bundler) and Bundler.rubygems.find_name('puppet_litmus').any?`

pmcmaw:
The force push after @DavidS approval is just a tidy up (removing commented out code) and squashing commits.

pmcmaw:
@ekohl we are planning on introducing this fix for MODULES-8778 introduced in #281. I would just like to ensure that this resolves your particular use case before progressing.

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -9,7 +9,7 @@
 require 'English'
 
 # dont load beaker if litmus is present
-require 'puppetlabs_spec_helper/tasks/beaker' unless Bundler.rubygems.find_name('puppet_litmus').any?
+require 'puppetlabs_spec_helper/tasks/beaker' unless Object.const_defined?(:Bundler) && Bundler.rubygems.find_name('puppet_litmus').any?
```
ekohl: How about:
```ruby
begin
  require 'beaker'
  require 'puppetlabs_spec_helper/tasks/beaker'
rescue LoadError
  # ignore
end
```

pmcmaw: @ekohl Our reason for attempting the above is we don't want the beaker tasks loaded at all if litmus is present. What are your thoughts on this?

```
begin
  require 'beaker'
  require 'puppetlabs_spec_helper/tasks/beaker' unless Object.const_defined?(:Bundler) && Bundler.rubygems.find_name('puppet_litmus').any?
rescue LoadError
  # ignore
end
```

ekohl: IMHO it's a bad idea. My reasoning is always that adding something should only add things, not remove them. This means that if users are starting to migrate some tests to litmus, they can't run anything on beaker anymore. By doing so, you block iterative migrations and force big bang migrations. That might hurt adoption.

Use cases for iterative migrations are complex tests. Right now litmus is lacking in the multi-node testing. However, users can experiment with some trivial tests in litmus (single node deployment) while keeping complex their tests in beaker.

Users who have migrated to litmus will remove beaker from their gemfiles and that way the tasks will also disappear.

pmcmaw: That is completely fair and all valid points.
My plan now is to close this PR and revert #281, then cut a release. The initial plan was to avoid loading both, to reduce confusion on showing both rake tasks from beaker and litmus and causing confusion. Thank you for your feedback on all of this it has been invaluable @ekohl 👍

pmcmaw:
Closing as per comments.

lib/puppetlabs_spec_helper/rake_tasks.rb
ekohl: You can still apply the trick to attempt loading beaker and if so, loading tasks. However, reverting and cutting a release is a quick fix that at least fixes previous behavior.

PR #288, opened 2019-04-15
(PDK-1047) Add rake task to inspect module for class & defined type definitions

rodjek:
This will eventually allow us to automatically generate skeleton tests
for untested classes & defined types.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -7,7 +7,10 @@ require 'puppetlabs_spec_helper/version'
 require 'puppetlabs_spec_helper/tasks/beaker'
 require 'puppetlabs_spec_helper/tasks/fixtures'
 require 'puppetlabs_spec_helper/tasks/check_symlinks'
+require 'puppetlabs_spec_helper/util'
 require 'English'
+
+extend PuppetlabsSpecHelper::Util
 
 # optional gems
 begin
@@ -377,5 +380,16 @@ if File.exist? locales_dir
     end
   rescue Gem::LoadError
     puts 'No gettext-setup gem found, skipping GettextSetup config initialization' if Rake.verbose == true
+  end
+end
+
+namespace :module do
+  desc 'Generate a JSON description of the module contents'
+  task :inspect do
+    raise 'This task does not work with Puppet <= 3' if puppet_3_or_older?
+
+    require 'puppetlabs_spec_helper/tasks/module_inspector'
+    data = PuppetlabsSpecHelper::Tasks::ModuleInspector.new.run
+    puts $stdout.isatty ? JSON.pretty_generate(data) : JSON.generate(data)
   end
 end
--- a/lib/puppetlabs_spec_helper/tasks/module_inspector.rb
+++ b/lib/puppetlabs_spec_helper/tasks/module_inspector.rb
@@ -1,0 +1,102 @@
+require 'set'
+require 'puppet'
+
+module PuppetlabsSpecHelper; end
+module PuppetlabsSpecHelper::Tasks; end
+
+class PuppetlabsSpecHelper::Tasks::ModuleInspector
+  attr_reader :path
+
+  def initialize(path = Dir.pwd)
+    @path = path
+  end
+
+  def run
+    {
+      host_class:   host_classes,
+      defined_type: defined_types,
+    }
+  end
+
+  private
+
+  def manifests
+    return [] unless File.directory?(File.join(path, 'manifests'))
+
+    Dir[File.join(path, 'manifests', '**', '*.pp')]
+  end
+
+  def manifest_definitions
+    @manifest_definitions ||= manifests.map { |r|
+      PuppetlabsSpecHelper::Tasks::ManifestInspector.new(r).definitions
+    }.flatten
+  end
+
+  def host_classes
+    remove_type_key(manifest_definitions.select { |r| r[:type] == :host_class })
+  end
+
+  def defined_types
+    remove_type_key(manifest_definitions.select { |r| r[:type] == :defined_type })
+  end
+
+  def remove_type_key(hash)
+    hash.map { |r| r.reject { |k, _| k == :type } }
+  end
+end
+
+class PuppetlabsSpecHelper::Tasks::ManifestInspector
+  attr_reader :path
+
+  DEFINITION_CLASSES = Set[
+    Puppet::Pops::Model::HostClassDefinition,
+    Puppet::Pops::Model::ResourceTypeDefinition,
+  ]
+
+  def initialize(path)
+    @path = path
+  end
+
+  def definitions
+    ast.definitions.find_all { |r| DEFINITION_CLASSES.include?(r.class) }.map do |obj|
+      obj_type = obj.class.to_s.split('::').last.gsub(%r{([a-z]+)([A-Z][a-z])}, '\1_\2').downcase
+      send("handle_#{obj_type}", obj)
+    end
+  end
+
+  private
+
+  def handle_host_class_definition(obj)
+    {
+      type:       :host_class,
+      file:       path,
+      name:       obj.name,
+      parameters: obj.parameters.map { |r| handle_parameter(r) },
+      parent:     obj.parent_class,
+    }
+  end
+
+  def handle_resource_type_definition(obj)
+    {
+      type:       :defined_type,
+      file:       path,
+      name:       obj.name,
+      parameters: obj.parameters.map { |r| handle_parameter(r) },
+    }
+  end
+
+  def handle_parameter(obj)
+    obj.name
+  end
+
+  def parser
+    Puppet::Pops::Parser::EvaluatingParser.new
+  end
+
+  def ast
+    @ast ||= begin
+               result = parser.parse_file(path)
+               result.respond_to?(:model) ? result.model : result
+             end
+  end
+end
--- a/lib/puppetlabs_spec_helper/util.rb
+++ b/lib/puppetlabs_spec_helper/util.rb
@@ -1,0 +1,9 @@
+require 'puppet'
+
+module PuppetlabsSpecHelper; end
+
+module PuppetlabsSpecHelper::Util
+  def puppet_3_or_older?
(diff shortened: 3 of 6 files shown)


lib/puppetlabs_spec_helper/util.rb
```diff
@@ -0,0 +1,9 @@
+require 'puppet'
+
+module PuppetlabsSpecHelper; end
+
+module PuppetlabsSpecHelper::Util
+  def puppet_3_or_older?
+    Puppet.version.split('.').first.to_i <= 3
```
glennsarti: Any reason you didn't use the semver stuff from `Gem::Version` ? e.g.

``` ruby
Gem::Version.new(Puppet.version) <  Gem::Version.new('4.0.0')
```

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -379,3 +382,14 @@
     puts 'No gettext-setup gem found, skipping GettextSetup config initialization' if Rake.verbose == true
   end
 end
+
+namespace :module do
+  desc 'Generate a JSON description of the module contents'
+  task :inspect do
+    raise 'This task does not work with Puppet <= 3' if puppet_3_or_older?
```
glennsarti: nit: Not a fan of `_or_older` but can't think of a better name :-(

Perhaps `puppet_3_or_less` or  `unless supports_puppet4_api?`

lib/puppetlabs_spec_helper/tasks/module_inspector.rb
```diff
@@ -0,0 +1,102 @@
...
+require 'puppet'
+
+module PuppetlabsSpecHelper; end
+module PuppetlabsSpecHelper::Tasks; end
+
+class PuppetlabsSpecHelper::Tasks::ModuleInspector
+  attr_reader :path
+
+  def initialize(path = Dir.pwd)
+    @path = path
```
glennsarti: Should the be a guard here if the `path` isn't the root of a module?

lib/puppetlabs_spec_helper/tasks/module_inspector.rb
```diff
@@ -0,0 +1,102 @@
+require 'set'
+require 'puppet'
+
+module PuppetlabsSpecHelper; end
+module PuppetlabsSpecHelper::Tasks; end
+
+class PuppetlabsSpecHelper::Tasks::ModuleInspector
```
glennsarti: nit: Should this class etc. be documented?

spec/unit/puppetlabs_spec_helper/tasks/module_inspect_spec.rb
```diff
@@ -0,0 +1,70 @@
+require 'spec_helper'
+
+describe 'rake module:inspect', type: :task, unless: puppet_3_or_older? do
+  let(:manifest_dir) { File.join(Dir.pwd, 'manifests') }
+  let(:no_results) { { 'host_class' => [], 'defined_type' => [] } }
+
+  before(:each) do
+    Dir.mkdir(manifest_dir)
```
glennsarti: Should this be cleaned up afterwards?

spec/unit/puppetlabs_spec_helper/tasks/module_inspect_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      File.open(File.join(manifest_dir, 'empty.pp'), 'w') { |fd| }
+    end
+
+    it 'runs without raising an error' do
+      expect { task.execute }.to output_json(no_results).to_stdout
+    end
+  end
+
+  context 'when there are manifests containing class and defined types' do
+    before(:each) do
```
glennsarti: Question: Instead of creating on the fly should we be using a fixtures directory instead?

spec/unit/puppetlabs_spec_helper/tasks/module_inspect_spec.rb
```diff
@@ -0,0 +1,70 @@
+require 'spec_helper'
+
+describe 'rake module:inspect', type: :task, unless: puppet_3_or_older? do
```
glennsarti: Should there be a matching test that it should error if puppet3 or older?

glennsarti: Also, a pattern we've adopted in modules land is to use `skip(...)` moreso than guards, that way we can see the tests are being skipped as opposed to missing entirely.

e.g. https://github.com/puppetlabs/puppetlabs-powershell/commit/62ef1a5b9fcccc2f9dcf56fa511bac513c6633e3

spec/unit/puppetlabs_spec_helper/tasks/module_inspect_spec.rb
glennsarti: Would save the creating and deleting of files on-the-fly

lib/puppetlabs_spec_helper/tasks/module_inspector.rb
```diff
@@ -0,0 +1,102 @@
...
+    Puppet::Pops::Model::HostClassDefinition,
+    Puppet::Pops::Model::ResourceTypeDefinition,
+  ]
+
+  def initialize(path)
+    @path = path
+  end
+
+  def definitions
+    ast.definitions.find_all { |r| DEFINITION_CLASSES.include?(r.class) }.map do |obj|
```
glennsarti: I assume if something "goes wrong" here the error is being surfaced via stderr and exit codes.  Is there a case to capture this and output in structured format? e.g.

``` json
{
  "error":  "THE WORLD IS CRASHING!"
  "file": "/some/path"
}
```

Should we return partial results?
Currently no tests for unhappy path of invalid manifests etc.

spec/unit/puppetlabs_spec_helper/tasks/module_inspect_spec.rb
rodjek: Nope, these tests use FakeFS (automatically included in `:type => :task` specs) which is an in memory filesystem

lib/puppetlabs_spec_helper/util.rb
rodjek: This is a bit lighter weight as I'm only concerned about the major version here which can be handled by a simple integer comparison, but I'm fine with updating to use `Gem::Version`s instead.

lib/puppetlabs_spec_helper/rake_tasks.rb
rodjek: I prefer `_or_older` rather than `_or_less` as it matches the wordage normally used when describing releases - "an older release" vs "a lesser release"

spec/unit/puppetlabs_spec_helper/tasks/module_inspect_spec.rb
rodjek: The task tests use an in memory filesystem, so the overhead of setting up and tearing down on the fly is negligible.

spec/unit/puppetlabs_spec_helper/tasks/module_inspect_spec.rb
rodjek: Yeah, my personal preference is to use the example group conditionals because I don't like having known and expected failure cases shown up as pending in the results. To me pending tests are more for "I know this thing is broken and I'm going to come back and fix it up later".

Regardless, this is a style decision that we should discuss as a team and decide on so that we can be consistent across the projects we're maintaining

spec/unit/puppetlabs_spec_helper/tasks/module_inspect_spec.rb
glennsarti: Ahh ok... hard to see that :-(  apologies

logicminds:
How much difference would this be to what https://github.com/nwops/puppet-retrospec is doing?  Seems like the wheel was reinvented again.

lib/puppetlabs_spec_helper/rake_tasks.rb
glennsarti: .. or perhaps be explicit. `support_inspection?`

glennsarti:
@logicminds Thanks for pointing to that project!  (I am not a lawyer), but I'm not sure the licensing of retrospec (GNU AGPL3) is compatible with license of this project (Apache 2)

logicminds:
I wasn't suggesting using retrospec as a library.  What I was saying is that retrospec already does the thing being built here.  It has been available since 2015 and I don't really know why this code is needed.  By building this your just asking for more technical debt and maintenance when your time could be spent elsewhere building other things in PDK. 

Test generators suffer from the problem that most people using them have no clue how to write or maintain the tests it creates.  Test Education is more important IMHO.

Also I would advise not adding this code here as it bloats what this gem. The test generator will eventually need to be migrated into a separate gem anyways.  Look how big retrospec-puppet is!  The approach here is what I started with 5 years ago and then abandon it because it wouldn't scale across nodes in the AST.  

If the goal is to really create test skeletons just tout the retrospec tool on the puppet  blog and help maintain the code to make it better.  Making retrospec work with PDK and the pdk templates would be an awesome addition.

glennsarti:
Apologies @logicminds.  I interpreted what you said the wrong way.

> Also I would advise not adding this code here as it bloats what this gem.

I would have to agree with this.  It does seem like PDK would be the better place for this functionality.  @rodjek @scotje  Thoughts?

rodjek:
> What I was saying is that retrospec already does the thing being built here.

The goal of PDK-1047 is not to implement retrospec style functionality. We do not want to inspect the contents of the class or defined types to generate extensive tests for all the contained resources.

> Test generators suffer from the problem that most people using them have no clue how to write or maintain the tests it creates. Test Education is more important IMHO.

The scope of this is simply to be able to examine a module during `pdk convert` and feed this information into the existing PDK object generators so that the *very minimal* template specs are generated, the same specs that are generated by running `pdk new class` or `pdk new defined_type`.

These generated tests are little more than `it { is_expected.to compile }`, which we've found is enough for most users.

chelnak:
👋 This has been open since 2019 and has had no further activity since.

I'm going to close this for house keeping, however the internal ticket will stick around for future reference.

PR #290, opened 2019-06-12
(MAINT) Removes old rubies and puppet versions

logicminds:
* Support for puppet 3.0 and ruby 1.9 is long gone
    so it has been removed from test matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,7 @@ os: linux
 dist: xenial
 language: ruby
 cache: bundler
-before_install:
-   bundle -v
+before_install: bundle -v
 script: env COVERAGE=yes bundle exec rake
 jobs:
   fast_finish: true
@@ -30,11 +29,5 @@ jobs:
     env: PUPPET_GEM_VERSION='~> 4.0'
   - rvm: '2.1'
     env: PUPPET_GEM_VERSION='~> 4.0'
-  - rvm: '2.1'
-    env: PUPPET_GEM_VERSION='~> 3.0'
-  - rvm: '2.0'
-    env: PUPPET_GEM_VERSION='~> 3.0'
-  - rvm: '1.9'
-    env: PUPPET_GEM_VERSION='~> 3.0'
 notifications:
   email: false


.travis.yml
```diff
@@ -1,35 +1,28 @@
...
-    env: PUPPET_GEM_VERSION='~> 4.0'
-  - rvm: '2.1'
-    env: PUPPET_GEM_VERSION='~> 4.0'
-  - rvm: '2.1'
-    env: PUPPET_GEM_VERSION='~> 3.0'
-  - rvm: '2.0'
-    env: PUPPET_GEM_VERSION='~> 3.0'
-  - rvm: '1.9'
-    env: PUPPET_GEM_VERSION='~> 3.0'
+    - rvm: "2.5"
```
bastelfreak: does it make sense to add ruby2.6 as well? (maybe in another PR)

logicminds: Yes that would be a good idea but I only removed the deprecated items in this request.

.travis.yml
```diff
@@ -1,35 +1,28 @@
...
+      env: PUPPET_GEM_VERSION='~> 6.0'
+    - rvm: "2.4"
+      env: PUPPET_GEM_VERSION='~> 6.0'
+    - rvm: "2.3"
+      env: PUPPET_GEM_VERSION='~> 6.0'
+    - rvm: "2.4"
+      env: PUPPET_GEM_VERSION='~> 5.0'
+    - rvm: "2.3"
+      env: PUPPET_GEM_VERSION='~> 5.0'
+    - rvm: "2.1"
```
glennsarti: This is interesting.

Puppet Agent 5.x (AIO) ships with Ruby 2.4. So I don't _think_ we need to test on 2.1?

That said, puppet gemspec for 5.x says it requires ruby 1.9.3+ .  The only reason I can think of for this is that puppet running on Puppet Server uses the old jruby which is 1.9.3 language compliant.

@rodjek  thoughts?

Ref: https://github.com/puppetlabs/puppet/blob/5.5.x/.gemspec#L21

glennsarti: We've been bitten previously with modules using "modern" ruby syntax and then it failing in Puppet Server because it's running old jruby (1.9.x syntax)

IIRC we've updated to jruby 9k but I can't remember exactly when

HOWEVER...

that's puppet server runtime, not necessarily puppet spec testing time which is where spec_helper resides

---

So perhaps just remove Ruby 2.1 for Puppet 5.x and this is a 👍 from me!

rodjek: JRuby9k was added as optional in puppet-server 5.x and not switched over as the default until puppet-server 6.x, so we should still keep testing Puppet 4.x and 5.x against Ruby 1.9.3.

logicminds:
How would puppetlabs_spec_helper code ever be run in a server?  These are development libraries for testing puppet code.

rodjek:
If my module contains Ruby that is executed on the master like types or functions then it's entirely feasible that I would want to run my tests against Ruby 1.9.3 or JRuby 1.7 in which case puppetlabs_spec_helper would be loaded.

daianamezdrea:
@logicminds, hello! Can you rebase this PR ? Thank you !

pmcmaw:
Thank you for your PR and such a quick rebase :-) 
Your contribution is really appreciated.

PR #291, opened 2019-06-27
Feature: metadata dependencies are auto installed

cardil:
Auto-Install of dependencies based on metadata.json info

Adding support to install modules from remote Puppet Forge, automatically, by reading contents of `metadata.json`. There are some configuration options added, like option to define `forge` address. Automatic installation is disabled by default to keep backward compatibility - in next major version it should be enabled by default.

Actual default value is:

```yaml
fixtures:
  metadata:
    autoinstall_dependencies: false
```

Additionally:
 * Adding installed module tree display after dependencies installation.
 * Added more tests, including acceptance, so that rake spec using this module will be asserted.


cardil:
@Cosaquee I'm notifying you as you always wanted this feature :wink:

cardil:
Due to [PUP-9813](https://tickets.puppetlabs.com/browse/PUP-9813) installation is rather slow. Merging PR https://github.com/puppetlabs/puppet/pull/7602 should provide a walkaround, until minitar performance can be enhanced.

logicminds:
#284 should speed up the downloads greatly.

cardil:
But it's not download that are a problem but extraction of archives. Especially large like stdlib.

Parallel execution might help, but not with this PR as it pass installation of modules and their dependencies to puppet.

daianamezdrea:
Hi @cardil, this looks nice! Can you rebase it, please ? Thank you! Cheers!

DavidS:
As much as I'd like to have the changes, I've tried rebasing this and didn't get anywhere as the core changes in fixtures download are clashing hard with the other changes that have went in.

If anyone gets around to resurrecting this code, please do open a new PR referencing this one.

PR #292, opened 2019-08-12
(MAINT) Fix initialize of Gettext call

ThoughtCrhyme:
`bundle exec rake changelog` was failing with `NoMethodError: undefined method `initialize_config' for GettextSetup:Module` changing to initialize fixed the error.


.idea/workspace.xml
```diff
@@ -0,0 +1,962 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bastelfreak: I think this file should not be in this commit?

ThoughtCrhyme: oops, yup

daianamezdrea:
@ThoughtCrhyme, this LGTM, thank you for your work! I'll merge this PR.

PR #294, opened 2019-10-03
Ignore plans folder and any subfolder

cyberious:


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -220,7 +220,7 @@
 PuppetSyntax.exclude_paths << 'spec/fixtures/**/*'
 PuppetSyntax.exclude_paths << 'pkg/**/*'
 PuppetSyntax.exclude_paths << 'vendor/**/*'
```
alexjfisher: Whilst you're at it, could you also add `.vendor/**/*`?  This is already in the puppet-lint ignore_paths, but missing here.

lionce:
Hello @cyberious ,

Would you please check the comments? We appreciate your help and support! Thanks!

Cheers!

lib/puppetlabs_spec_helper/rake_tasks.rb
cyberious: Hi @alexjfisher, I see your point, why would we have Puppet Code in a `.vendor` repo, every time I have used a `.vendor` it has been for Ruby Gems. I am fine making this change but for Puppet Syntax I am scratching my head as to why we would hide modules in `.vendor`

cyberious:
@lionce commented above and added change

lionce:
Hello @cyberious ,

Thank you! Appreciate your help and patience.

Cheers!

PR #310, opened 2020-05-09
(MODULES-10658) Require puppet-syntax 3.x

bastelfreak:


puppetlabs_spec_helper.gemspec
```diff
@@ -22,7 +22,7 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency 'mocha', '~> 1.0'
   spec.add_runtime_dependency 'puppet-lint', '~> 2.0'
-  spec.add_runtime_dependency 'puppet-syntax', '~> 2.0'
+  spec.add_runtime_dependency 'puppet-syntax', '>= 2.0'
```
alexjfisher, line 25: Would `~> 3.0` be better?  Just in case we do something crazy with puppet-syntax 4 in the future?

bastelfreak: ah agree

PR #311, opened 2020-05-09
Add Ruby 2.6/Puuppet6 to CI matrix

bastelfreak:
Ruby 2.6 isn't even the newest version, but many people use it locally
to run their tests. Based on that I think it makes sense to add it to
the CI matrix.


david22swan:
@bastelfreak Thanks for the pr :)

PR #330, opened 2021-01-21
Restructure PuppetLint rake tasks so they can be configurable

nmaludy:
Currently the spec helper configures `PuppetLint` in such a way that it's not easy to add extra directories or disable checks for a local module. Basically, you have to completely redefine the Rake task exactly how `puppetlabs_spec_helper` does it today, which is not idea.

Instead, this changes over to using the `PuppetLint.configuration` class to set the global options, this way things like `ignore_paths` can be appended to easily in a `spec_helper_local.rb` like so:

```
PuppetLint.configuration.ignore_paths << 'modules/**/*'
```

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -172,52 +172,34 @@ require 'puppet-lint/tasks/puppet-lint'
 # Must clear as it will not override the existing puppet-lint rake task since we require to import for
 # the PuppetLint::RakeTask
 Rake::Task[:lint].clear
-# Relative is not able to be set within the context of PuppetLint::RakeTask
+# Utilize PuppetLint global configuration so that these settings can be tweaked by
+# spec_helper.rb in an individual module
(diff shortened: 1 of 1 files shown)


tphoney (approved):
nice fix 👍

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -172,52 +172,34 @@
 # Must clear as it will not override the existing puppet-lint rake task since we require to import for
 # the PuppetLint::RakeTask
 Rake::Task[:lint].clear
-# Relative is not able to be set within the context of PuppetLint::RakeTask
+# Utilize PuppetLint global configuration so that these settings can be tweaked by
+# spec_helper.rb in an individual module
 PuppetLint.configuration.relative = true
-PuppetLint::RakeTask.new(:lint) do |config|
-  config.fail_on_warnings = true
```
ekohl, line 178: This setting was lost in the refactor. I opened https://github.com/puppetlabs/puppetlabs_spec_helper/pull/347 to fix the regression.

PR #332, opened 2021-02-09
dropping rubies before 2.4

DavidS:
Since dropping the gettext-setup gem and rake task in #331 is already a breaking change, might as well go ahead and drop rubies from the before times and require ruby2.4 or newer from now on.


DavidS:
renamed the PR - this needs to get merged and then I can update the changelog to accurately depict the changes in this release.

michaeltlombardi (approved):

PR #337, opened 2021-05-21
Reuse fixture location helper

ekohl:
Rather than hardcoding the location, the helper is now used to reduce duplication. While unlikely to ever change, it helps consistency.


da-ar:
Thanks for the contribution, however, this minor change has a large knock on effect to the tests.

PR #338, opened 2021-05-21
Remove beaker integration

ekohl:
Today there's both beaker and litmus. The spec helper shouldn't be opinionated. For beaker integration, I'd prefer to make this easy to consume.

beaker-rspec[1] can provide useful helpers. There is already beaker-rspec/rake_task[2] which feels like a more native place.

In Voxpupuli there's also voxpupuli-acceptance[3] which is a potential place.

[1] https://github.com/voxpupuli/beaker-rspec
[2] https://github.com/voxpupuli/beaker-rspec/blob/master/lib/beaker-rspec/rake_task.rb
[3] https://github.com/voxpupuli/voxpupuli-acceptance


da-ar:
Hi @ekohl thanks for raising this. We're looking into whether it's possibile to make thi beaker code an optional include so that this doesnt affect existing users who may make use of the functionality.

ekohl:
Isn't that easy? I opened https://github.com/puppetlabs/puppetlabs_spec_helper/pull/339 but right now I can't easily test it (upgraded to Fedora 34 with Ruby 3).

jpogran (approved):
Thank you for your contribution! Apologies on the delay, my fault entirely

ekohl:
Does this mean the next version is a major version bump?

PR #339, opened 2021-05-25
Only load beaker tasks if beaker-rspec is present

ekohl:
Not everyone is using beaker and there's no point in adding those rake tasks. This tries to load beaker-rspec. If that works, we can assume Beaker is present.

Currently a test because I haven't tested it. Just showing it as a response to https://github.com/puppetlabs/puppetlabs_spec_helper/pull/338#issuecomment-847115161.


ekohl:
Obsolete now that https://github.com/puppetlabs/puppetlabs_spec_helper/pull/338 is merged.

PR #346, opened 2021-11-11
Fail if git ls-files failed

ekohl:
WHen running tests I noticed this:

    fatal: ls-files -i must be used with either -o or -c

However, rake proceeded so it wasn't failing.

This isn't a complete PR, but I wanted to at least make this obvious and so I wouldn't forget.


bastelfreak:
For me this also always fails with `fatal: ls-files -i must be used with either -o or -c` and an exitcode of 128. I think the command should be `git ls-files --ignored --exclude-standard --cached`

ekohl:
I included https://github.com/puppetlabs/puppetlabs_spec_helper/pull/351 now.

chelnak:
Looking like this will go in rather than #351.

Again, will need rebasing as the changes come in.

chelnak:
On a second look, this can actually go in now if you are happy @ekohl ?

ekohl:
I did a trivial rebase and would be happy if this was merged & released.

PR #347, opened 2021-11-17
Configure puppet-lint to fail on warnings again

ekohl:
In a4bed8a2af6734744715178fd5f2b11e8aa7ded1 the checks were restructured to be reconfigurable but the `fail_on_warnings` setting was lost. This means the rake task isn't failing even if there are warnings.

Fixes: a4bed8a2af6734744715178fd5f2b11e8aa7ded1

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -171,6 +171,7 @@ require 'puppet-lint/tasks/puppet-lint'
 # Must clear as it will not override the existing puppet-lint rake task since we require to import for
 # the PuppetLint::RakeTask
 Rake::Task[:lint].clear
+PuppetLint.configuration.fail_on_warnings = true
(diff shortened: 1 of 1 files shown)


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -171,6 +171,7 @@
 # Must clear as it will not override the existing puppet-lint rake task since we require to import for
 # the PuppetLint::RakeTask
 Rake::Task[:lint].clear
+PuppetLint.configuration.fail_on_warnings = true
```
bastelfreak, line 174: I'm :100: to enable this. But I've no idea how this relates to
* https://github.com/puppetlabs/pdk-templates/blob/main/moduleroot/Rakefile.erb#L70
* https://github.com/puppetlabs/pdk-templates/blob/main/moduleroot/.puppet-lint.rc.erb#L18

michaeltlombardi (approved):
LGTM, thanks for catching this.

PR #348, opened 2022-02-15
Drop outdated future parser support

binford2k:
This is outdated and no longer supported by [puppet-syntax](https://github.com/voxpupuli/puppet-syntax/pull/120)


bastelfreak (approved):
looks good to me. Is that a breaking change? or di we say that such old puppet versions arent supported by latest puppetlabs_spec_helper anyways?

binford2k:
It would be a breaking change, except that we already accidentally broke when [`puppet-syntax` released 3.0](https://github.com/voxpupuli/puppet-syntax/blob/master/CHANGELOG.md#v300-2020-05-09)

chelnak:
Thanks @binford2k

chelnak (approved):

PR #352, opened 2022-06-01
Run the `strings:validate:reference` task during `validate`

smortex:
If the documentation is not up-to-date or is broken, we want the validate task to fail so that we can more quickly detect inconsistencies and fix them early.

This was initially part of https://github.com/puppetlabs/puppetlabs-stdlib/pull/1239 but all projects can benefit from it.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -228,6 +228,14 @@ task :validate do
       warn 'Skipping metadata validation; the metadata-json-lint gem was not found'
     end
   end
+
+  if File.exist?('REFERENCE.md')
+    if Rake::Task.task_defined?('strings:validate:reference')
+      Rake::Task['strings:validate:reference'].invoke
+    else
+      warn 'Skipping reference documentation validation; the puppet-strings gem was not found'
+    end
+  end
 end
 
 task :metadata do


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -31,6 +31,10 @@
 
 begin
   require 'puppet-strings/tasks'
+  # strings:validate:reference would raise an error if REFERENCE.md is not
+  # up-to-date, but we only want to be sure that all documentation is in place
+  # and ensure it is kept up-to-date.
+  Rake::Task[:validate].enhance ['strings:generate:reference']
```
bastelfreak, line 37: this ahould only be run if the file exists. theREFERENCE.md is unuseable if params are documented in another way. at voy pupuli we fixed it like this: https://github.com/voxpupuli/voxpupuli-release/blob/master/lib/voxpupuli/release/rake_tasks.rb#L18

smortex: Make sense, I updated the PR accordingly.

bastelfreak (approved):

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -228,6 +228,10 @@
       warn 'Skipping metadata validation; the metadata-json-lint gem was not found'
     end
   end
+
+  if File.exist?('REFERENCE.md')
+    Rake::Task['strings:validate:reference'].invoke
```
ekohl, line 233: The task may not be defined because there is graceful handling when the gem is not present:
https://github.com/puppetlabs/puppetlabs_spec_helper/blob/ac4945e2ccaa20a6868d29051d6e701aff73d3d7/lib/puppetlabs_spec_helper/rake_tasks.rb#L32-L36

I'd check if the task exists:
```suggestion
    if Rake::Task.task_defined?(:metadata_lint)
      Rake::Task['strings:validate:reference'].invoke
    else
      warn 'Skipping reference documentation validation; the puppet-strings gem was not found'
    end
```

ekohl (changes requested):

lib/puppetlabs_spec_helper/rake_tasks.rb
smortex: Makes sense! Updated to match the actual task we want to invoke.

chelnak (approved):

ekohl (approved):

david22swan (approved):

PR #353, opened 2022-08-05
Add rspec-github integration

ekohl:
This adds rspec-github as a dependency and at runtime it is optionally configured. The detection is the same as for Rubocop.

In https://github.com/voxpupuli/puppet-example/pull/22 I've showcased that this works:
Screen Shot 2022-08-05 at 15 44 53

### DIFF
--- a/puppetlabs_spec_helper.gemspec
+++ b/puppetlabs_spec_helper.gemspec
@@ -32,5 +32,6 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'pathspec', '~> 1.0'
   spec.add_runtime_dependency 'puppet-lint', '~> 3.0'
   spec.add_runtime_dependency 'puppet-syntax', '~> 3.0'
+  spec.add_runtime_dependency 'rspec-github', '~> 2.0'
   spec.add_runtime_dependency 'rspec-puppet', '~> 2.0'
 end
--- a/lib/puppetlabs_spec_helper/module_spec_helper.rb
+++ b/lib/puppetlabs_spec_helper/module_spec_helper.rb
@@ -62,6 +62,10 @@ components.flatten.each do |d|
 end
 
 RSpec.configure do |c|
+  if ENV['GITHUB_ACTIONS'] == 'true'
+    c.formatter = 'RSpec::Github::Formatter'
+  end
+
   c.environmentpath = spec_path if Puppet.version.to_f >= 4.0
   c.module_path = module_path
   c.manifest_dir = File.join(fixture_path, 'manifests')


puppetlabs_spec_helper.gemspec
```diff
@@ -26,6 +26,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'pathspec', '>= 0.2.1', '< 1.1.0'
   spec.add_runtime_dependency 'puppet-lint', '~> 2.0'
   spec.add_runtime_dependency 'puppet-syntax', ['>= 2.0', '< 4']
+  spec.add_runtime_dependency 'rspec-github', '~> 2.0'
```
ekohl, line 29: As an alternative to making it a hard dependency, it can also be made a soft dependency:
```ruby
RSpec.configure do |c|
  if ENV['GITHUB_ACTIONS'] == 'true'
    begin
      c.formatter = 'RSpec::Github::Formatter'
    rescue LoadError
      # rspec-github not present
    end
  end
```

jhoblitt: Does it really matter if it is a hard dep?  As long as there is a flag to turn it off for other CI systems, I don't see the gem being pulled as much of a problem.

jhoblitt: If it isn't a hard dep, that would mean that `rspec-github` would have to be manually added to (literally) 100s of repos.

ekohl: > Does it really matter if it is a hard dep? As long as there is a flag to turn it off for other CI systems, I don't see the gem being pulled as much of a problem.

Every additional gem is a potential problem. Just wanted to highlight it.

> If it isn't a hard dep, that would mean that `rspec-github` would have to be manually added to (literally) 100s of repos.

That was also my consideration. This looked easier for the average case.

bastelfreak (approved):
I had a discussion with @binford2k yesterday. I think this is a sane and useful patch. I also think rspec-github should be added as hard dependency, not a soft one. The patch was tested at Vox pupuli and it's working fine.

chelnak:
@ekohl I love this.

There are going to be a few more updates to the repo so this PR will need rebasing soon.. (unless you want to keep rebasing as we merge changes).

chelnak:
@ekohl @bastelfreak The bulk of work has been merged to main now. Would it be possible for one of you to rebase this and resolve the conflict? Once that is sorted i'll get it merged & it will be in the next release.

chelnak (approved):
Exciting, thank you for this.

PR #356, opened 2022-10-24
(CONT-236) Workflow updates

chelnak:
This PR adds the common set of reusable actions and removes travis.


GSPatton (approved):
all looks good. Tests are complaining about rubocop.yml though... I don't see the error being caused by this PR though

PR #357, opened 2022-10-24
(CONT-237) Rubocop and Rakefile updates

chelnak:
This PR updates the rubocop rules so that they match rules that are used on other Gems.

Additionally we also clean up the Rakefile.


GSPatton (approved):
looks good

PR #358, opened 2022-10-24
(CONT-237) Bump minimum Ruby version requirement

chelnak:
Primarily, this commit bumps the minimum required ruby version to 2.7.

Along with this change a number of dependencies have also been bumped so that they are inline with other gems.


GSPatton (approved):
looks good

Issue #372, opened 2023-01-23
Ruby 2.7+ required but Puppet 6 doesn't run on Ruby 2.7


ekohl:
AFAIK Puppet 6 doesn't run on Ruby 2.7. I'd expect either Ruby 2.5 + Puppet 6 or Ruby 2.7 + Puppet 7. However, given Puppet 6 is not yet EOL I'd prefer to have Ruby 2.5 as the lower bound.

_Originally posted by @ekohl in https://github.com/puppetlabs/puppetlabs_spec_helper/pull/364#discussion_r1084092481_

chelnak:
Yeah you are right..

Lets get that fixed.. hopefully there are no major dependency issues.

PR #373, opened 2023-01-23
(GH-372) Reset min Ruby requirement

chelnak:
Prior to this change the required Ruby version was set to greater than or equal to 2.7. This meant that the spec_helpers gem would not run on systems with Ruby 2.5.

This change updates the required Ruby version so that Rubies 2.5 and above are compatibel.

Additionally the following gems were downgraded for compatibility:

* pathspec is now ~> 0.2
* puppet-lint is now ~> 2.5.2

### DIFF
--- a/puppetlabs_spec_helper.gemspec
+++ b/puppetlabs_spec_helper.gemspec
@@ -26,11 +26,11 @@ Gem::Specification.new do |spec|
   spec.executables = Dir['bin/**/*'].map { |f| File.basename(f) }
   spec.require_paths = ['lib']
 
-  spec.required_ruby_version = Gem::Requirement.new('>= 2.7')
+  spec.required_ruby_version = Gem::Requirement.new('>= 2.5')
 
   spec.add_runtime_dependency 'mocha', '~> 1.0'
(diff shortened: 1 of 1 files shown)


GSPatton (approved):
looks good to me

ekohl:
Why didn't CI run on this? I don't see any checks. https://github.com/puppetlabs/puppetlabs_spec_helper/blob/main/.github/workflows/ci.yml (or https://github.com/puppetlabs/cat-github-actions/blob/main/.github/workflows/gem_ci.yml) also doesn't appear to run in a matrix build so how do you know it actually works?

ekohl:
Ah yes, combining cron and PR testing in the same flow. I've been hit by that too. I recommend adding required checks to avoid it. In gha-puppet we have a dummy job that depends on all others, so you can simply make that required instead of the whole matrix: https://github.com/voxpupuli/gha-puppet/blob/a8fea30a9fcf14fd4cc9071c839e5a55d5dbab83/.github/workflows/beaker.yml#L140-L149

chelnak:
Nice I'll take a look

PR #374, opened 2023-01-23
(MAINT) Ensure each job has a test matrix

chelnak:


ekohl (commented):
Can we also run on Ruby 3, 3.1? I'm not sure if it's 3.2 compatible yet, but that would be great to know as well. I always run on my developer machine with a simple `bundle install` and Fedora 36 and 37 have Ruby 3.

chelnak:
We can try! I

chelnak:
@ekohl on the topic of 3.2.. after some thought I'm not going to be adding in this PR.

Appreciate that might be frustrating.. however this is something we will be rolling out across the board very soon.

chelnak:
> I don't mind about 3.2 (since I know it can break a lot), but at least 3.0 and 3.1 would be nice. I already run it with 3.1 all the time so that should be safe.

done

LukasAud (approved):

PR #379, opened 2023-01-23
(CONT-515) Fix uninitialized constant error

chelnak:
This PR fixes the following error:

```
NameError: uninitialized constant PuppetlabsSpecHelper
/home/runner/work/pdk/pdk/vendor/bundle/ruby/2.7.0/gems/puppetlabs_spec_helper-5.0.1/lib/puppetlabs_spec_helper/tasks/fixtures.rb:8:in `<top (required)>'
/home/runner/work/pdk/pdk/Rakefile:3:in `require'
/home/runner/work/pdk/pdk/Rakefile:3:in `<top (required)>'
/home/runner/work/pdk/pdk/vendor/bundle/ruby/2.7.0/gems/rake-12.3.3/exe/rake:27:in `<top (required)>'
/opt/hostedtoolcache/Ruby/2.7.7/x[6](https://github.com/puppetlabs/pdk/actions/runs/3989741069/jobs/6842538417#step:7:7)4/bin/bundle:23:in `load'
/opt/hostedtoolcache/Ruby/2.[7](https://github.com/puppetlabs/pdk/actions/runs/3989741069/jobs/6842538417#step:7:8).7/x64/bin/bundle:23:in `<main>'
```

### DIFF
--- a/lib/puppetlabs_spec_helper/tasks/fixtures.rb
+++ b/lib/puppetlabs_spec_helper/tasks/fixtures.rb
@@ -4,406 +4,407 @@ require 'yaml'
 require 'open3'
 require 'json'
 
-# Top level namespace for spec helper tasks.
-module PuppetlabsSpecHelper::Tasks end
-
-# Helpers for workfing with fixtures.
-module PuppetlabsSpecHelper::Tasks::FixtureHelpers
-  # This is a helper for the self-symlink entry of fixtures.yml
-  def source_dir
-    Dir.pwd
-  end
-
-  # @return [String] - the name of current module
-  def module_name
-    raise ArgumentError unless File.file?('metadata.json') && File.readable?('metadata.json')
-
-    metadata = JSON.parse(File.read('metadata.json'))
-    metadata_name = metadata.fetch('name', nil) || ''
-
-    raise ArgumentError if metadata_name.empty?
-
-    metadata_name.split('-').last
-  rescue JSON::ParserError, ArgumentError
-    File.basename(Dir.pwd).split('-').last
-  end
-
-  def module_version(path)
-    metadata_path = File.join(path, 'metadata.json')
-    raise ArgumentError unless File.file?(metadata_path) && File.readable?(metadata_path)
-
-    metadata = JSON.parse(File.read(metadata_path))
-    metadata.fetch('version', nil) || '0.0.1'
-  rescue JSON::ParserError, ArgumentError
-    logger.warn "Failed to find module version at path #{path}"
-    '0.0.1'
-  end
-
-  # @return [Hash] - returns a hash of all the fixture repositories
-  # @example
-  # {"puppetlabs-stdlib"=>{"target"=>"https://gitlab.com/puppetlabs/puppet-stdlib.git",
-  # "ref"=>nil, "branch"=>"main", "scm"=>nil,
-  # }}
-  def repositories
-    @repositories ||= fixtures('repositories') || {}
-  end
-
-  # @return [Hash] - returns a hash of all the fixture forge modules
-  # @example
-  # {"puppetlabs-stdlib"=>{"target"=>"spec/fixtures/modules/stdlib",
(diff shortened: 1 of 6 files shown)


lib/puppetlabs_spec_helper/tasks/fixtures.rb
```diff
@@ -4,6 +4,9 @@
 require 'open3'
 require 'json'
 
+# Top level namespace for spechelper.
```
ekohl, line 7: Or you can use the common Ruby style:
```ruby
module PuppetlabsSpecHelper
  module Tasks
    module FixtureHelpers
      # ...
    end
  end
end
```

ekohl: https://rubystyle.guide/#namespace-definition

chelnak: Thanks for the suggestion. 

If nested modules have to exist, I actually prefer this way too but was battling with rubocop at the time.

Have updated accordingly.

lib/puppetlabs_spec_helper/tasks/fixtures.rb
```diff
@@ -1,409 +1,414 @@
...
-  # @return [String] - the name of current module
-  def module_name
-    raise ArgumentError unless File.file?('metadata.json') && File.readable?('metadata.json')
-
-    metadata = JSON.parse(File.read('metadata.json'))
-    metadata_name = metadata.fetch('name', nil) || ''
-
-    raise ArgumentError if metadata_name.empty?
+module PuppetlabsSpecHelper
+  # The Tasks module.
```
ekohl, line 11: This comment feels a bit redundant.

lib/puppetlabs_spec_helper/tasks/fixtures.rb
```diff
@@ -1,409 +1,414 @@
...
+# rubocop:disable Layout/ClassAndModuleChildren
+
 require 'yaml'
 require 'open3'
 require 'json'
 
-# Top level namespace for spec helper tasks.
-module PuppetlabsSpecHelper::Tasks end
-
 # Helpers for workfing with fixtures.
```
ekohl, line 9:
```suggestion
# Helpers for working with fixtures.
```

lib/puppetlabs_spec_helper/tasks/fixtures.rb
```diff
@@ -1,409 +1,414 @@
 # frozen_string_literal: true
 
+# rubocop:disable Layout/ClassAndModuleChildren
```
ekohl, line 3: I think Rubocop does this weird thing where you also have to enable it again. I'm not sure if that's per-file or global though. Do you know? Should we disable this in the Rubocop config instead?

bastelfreak (approved):
looks good to me. also thousand times more readable without the whitespace: https://github.com/puppetlabs/puppetlabs_spec_helper/pull/379/files?diff=unified&w=1

lib/puppetlabs_spec_helper/tasks/fixtures.rb
chelnak: I feel like it's reasonable to move this to the config. It's frustrating at best of times so to have it removed would be good.

lib/puppetlabs_spec_helper/tasks/fixtures.rb
chelnak: Another rubocopism.

I might actually spend a few minutes and see if there is a nicer way to do this.

ekohl: Yes, technically you can document modules. If you use rdoc or yardoc it will create some documentation. However, I always find no documentation better than some placeholder to please a lint check. You can disable the cop and I'd recommend that.

david22swan (approved):

ekohl (approved):

PR #381, opened 2023-01-25
Use select.last on a hash

ekohl:
In 24d92da1ce27de1fcd532dfaac440233429b7a0f this was changed to reverse.find but that's not defined on Hash. The Rubocop suggestion is actually false, so it's disabled.

Fixes: 24d92da1ce27de1fcd532dfaac440233429b7a0f

### DIFF
--- a/lib/puppetlabs_spec_helper/tasks/fixtures.rb
+++ b/lib/puppetlabs_spec_helper/tasks/fixtures.rb
@@ -317,7 +317,8 @@ module PuppetlabsSpecHelper
             end
           else
             # the last thread started should be the longest wait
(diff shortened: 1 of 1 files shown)


bastelfreak (approved):

chelnak:
Ah just noticed this. 

Looks like my PR (#382) has gone in. It just resets back to the original behaviour.

On the topic though, a TIL moment for me. I did not realise that find_all and select had subtle differences.

https://medium.com/@elizabethkosowski/ruby-find-all-vs-select-whats-the-deal-d0e2c8e7c5cb

ekohl:
I think you're right: it actually isn't the same and breaks.

PR #390, opened 2023-04-13
(CONT-807) Add Ruby 3.2 support

GSPatton:
This PR adds Ruby 3.2 support in preparation for Puppet 8 and removes Ruby 2.5 and Puppet 6 support


README.md
```diff
@@ -257,7 +257,7 @@ fixtures:
       ref: "2.6.0"
 ```
 
-Only install the `yumrepo_core` module when testing against Puppet 6 or greater:
```
david22swan, line 260: Is this really necessary now that everything below Puppet 7 is unsupported?

david22swan (commented):
One small comment

README.md
GSPatton: i originally removed all of this but thought it still had a place since it still seems relevant to >6. Seems now that this module should always be installed since all supported puppets are >6

david22swan (approved):

PR #401, opened 2023-07-12
(maint) Relax Ruby requirement

mhashizume:
Prior to this commit, pupeptlabs_spec_helper required Ruby >= 2.7 in its Gemspec.

Puppetserver 7, which is expected to be supported as part of PE 2021.7.z LTS through 2025, uses JRuby 9.3.y.z which targets compatibility with Ruby 2.6 and uses that in its RUBY_VERSION.

Also, Puppet 7 lists its required Ruby as >= 2.5 in the Gemspec.

This commit relaxes the gemspec's Ruby requirements from >= 2.7 to >= 2.5 to allow full compatibility with Puppet and Puppetserver 7.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -12,7 +12,7 @@ AllCops:
     - vendor/bundle/**/*
   NewCops: enable
   SuggestExtensions: false
-  TargetRubyVersion: '2.7'
+  TargetRubyVersion: '2.5'
 
 # Disabled
 Style/ClassAndModuleChildren:
--- a/puppetlabs_spec_helper.gemspec
+++ b/puppetlabs_spec_helper.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
   spec.executables = Dir['bin/**/*'].map { |f| File.basename(f) }
   spec.require_paths = ['lib']
 
-  spec.required_ruby_version = Gem::Requirement.new('>= 2.7')
+  spec.required_ruby_version = Gem::Requirement.new('>= 2.5')
 
   spec.add_runtime_dependency 'mocha', '~> 1.0'
   spec.add_runtime_dependency 'pathspec', '>= 0.2', '< 2.0.0'


mhashizume:
I ran into issues with this originally when Dependabot attempted to bump spec_helper in the resource_api: https://github.com/puppetlabs/puppet-resource_api/pull/313

davidsandilands:
Hey @mhashizume I wanted to check why we want this change? Is there an issue it causes you directly or something that can't be delivered?

There are a number of technical issues expanding out the supported ruby versions so we want to understand the need to change this.

mhashizume:
Hi @davidsandilands , this is not directly a blocker or problem for me at the moment, but a few thoughts:

- Requiring Ruby 2.7 breaks compatibility with JRuby on the Puppet 7 platform for reasons I mentioned in my commit message.
- It was unexpected to me for this gem to have a stricter Ruby requirement than the version of Puppet it supports.
- It looks like the current PDK template will set the requirement for this gem to `~> 6.0` so as users update their modules, they could encounter surprise compatibility issues.

That being said, as I mentioned this is not an issue for me personally right now. I just thought I'd put up this PR in case other folks hit this issue directly or the CAT team anticipates that this could be a larger issue as people move over to >= 6.0.

bastelfreak:
> Requiring Ruby 2.7 breaks compatibility with JRuby on the Puppet 7 platform

Yes, but why does it matter? the gem isnt ment to be used on old jruby. We bumped to 2.7 because thats what Puppet Agent 7 uses, also ruby 2.5 is EoL.

davidsandilands:
@mhashizume following conversation with with the team, we have decided as this is not a blocking issue, so we are not going to go forward with relaxing the pins. The reason being relaxing the pins suggest we support modules on the older versions of Ruby (which we don’t, they are just too old and it is impossible to support everything). That does not mean that they aren’t compatible we just don’t feel it is feasible or makes sense to support and test on versions of ruby that have been EOL for a long period of time.

If this position does change and it becomes blocking please contact us and we can work out best approach with you.

PR #406, opened 2023-09-05
README.md: replace git:// with https://

kenyon:
GitHub disabled the git protocol, so git:// should be replaced with
https://. https://github.blog/changelog/2022-03-15-removed-unencrypted-git-protocol-and-certain-ssh-keys/


ekohl (approved):
I think technically `.git` is not needed, but doesn't hurt either.

bastelfreak (approved):

PR #417, opened 2023-11-22
(CAT-1603) - Revert dependencies

jordanbreen28:
this PR reverts the changes used to pinpoint to the newly proposed puppetlabs gems.
These gems are no longer being published.

## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.


bastelfreak:
This isn't a `git revert` of the original commit, right? Would that be an option?

jordanbreen28:
> This isn't a `git revert` of the original commit, right? Would that be an option?

thanks @bastelfreak 
have opted to revert the original commits instead

LukasAud (approved):
LGTM

PR #424, opened 2024-01-16
(maint) - Make codecov gem support optional - changes to spec:simplecov rake task

jordanbreen28:
## Summary

The codecov gem has been deprecated and unsupported for quite some time now. (2+ years).
This means we are restricted to outdated versions < 0.22 of simplecov which include bugs with ruby 3.2, and naturally wont be receiving further updates. (See [here](https://github.com/simplecov-ruby/simplecov/issues/1003))

We should update the module_spec_helper file to only include this gem if installed by the user, else update documentation to guide users in how to now upload the coverage reports to codecov.io.

## Additional Context
See codecov repo https://github.com/codecov/codecov-ruby

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,2 +1,2 @@
 # Setting ownership to the tooling team and Tim
-* @puppetlabs/tooling @bastelfreak
+* @puppetlabs/devx @bastelfreak


ekohl (approved):

ekohl:
Slight nit: I'd rephrase the title as "Make codecov support optional" instead of "Add". Add implies it wasn't there before, but it was. Also, there's a typo in `spec:simplecov`.

jordanbreen28:
ah! derp. @ekohl Thanks for all your help on this one!

lib/puppetlabs_spec_helper/module_spec_helper.rb
```diff
@@ -20,35 +20,42 @@ def verify_contents(subject, title, expected_lines)
 module_path = File.join(fixture_path, 'modules')
 
 module_path = [module_path, env_module_path].join(File::PATH_SEPARATOR) if env_module_path
+
 if ENV['SIMPLECOV'] == 'yes'
```
bastelfreak, line 24: Is that if statement really required when you later on already catch the LoadError?

ekohl: This is a way to force it to be fatal. That way you can detect misconfigurations (missing in Gemfile). You could also argue you don't want coverage all the time (because it's slower, distracts you, etc). If you want to change it, I'd move that to a separate discussion and merge this now. It's now neatly contained to just making codecov optional.

LukasAud (approved):
LGTM

PR #439, opened 2024-02-15
task to check if puppet resources have a spec described

h0tw1r3:
Anyone interested in this?
Request for comment / improvements / toss it out?

Example:
```
$ pdk bundle exec rake spec_described[90]
pdk (INFO): Using Ruby 3.2.2
pdk (INFO): Using Puppet 8.4.0
===> rake: spec_described
Spec described coverage: 92.9%
  * 4 defines: 100.0%
  * 21 classes: 95.2%
    missing:
    - puppetdb::something
    unknown:
    - something in spec/classes/whoknows_spec.rb
  * 1 type_alias: 0.0%
    missing:
    - Puppetdb::Ttl
  * 2 functions: 100.0%
  ```

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -8,6 +8,7 @@ require 'pathname'
 require 'puppetlabs_spec_helper/version'
 require 'puppetlabs_spec_helper/tasks/fixtures'
 require 'puppetlabs_spec_helper/tasks/check_symlinks'
+require 'puppetlabs_spec_helper/tasks/spec_described'
 require 'English'
 
 # optional gems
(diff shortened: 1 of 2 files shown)


ekohl (commented):
I like the idea, but I don't want to test my private classes (like `$module::params`). Sadly you can only declare a class as private using puppet-strings so that may be hard with your existing code.

Overall puppet-strings can already parse and extract the various types, which is probably better than using regular expressions. Any thoughts on using puppet-strings to generate a list of all resources?

h0tw1r3:
> Overall puppet-strings can already parse and extract the various types, which is probably better than using regular expressions. Any thoughts on using puppet-strings to generate a list of all resources?

Thanks for the feedback. I'll check it out, cheers!

gavindidrichsen:
Hi @h0tw1r3.  We all like this on devx, so kicked off a discussion https://github.com/puppetlabs/puppetlabs_spec_helper/discussions/442

bastelfreak:
I like this and share the same opinion as @ekohl . I think this would be a helpful addition to puppetlabs_spec_helper.

PR #445, opened 2024-04-04
puppet-syntax: Ensure we are using 4.1.1 or newer

bastelfreak:
4.1.0 had a bug in the Hiera syntax validation. This is fixed in 4.1.1. See https://github.com/voxpupuli/puppet-syntax/pull/171 for details.

## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.


bastelfreak:
This project has a `bug` and `bugfix` label, I'm not sure which one is correct :thinking: . Could one of them get deleted?

jordanbreen28 (approved):
Thanks for this @bastelfreak!
RE the labels, we have bug for bug reports through issues and bugfix for prs. In the context of our changelog generator i believe bug & bugfix are inter-changeable, so doesn't matter much which one you use in the grand scheme of things. I'll update to bugfix for consistency sake.

PR #446, opened 2024-04-09
parallel_spec_standalone: use --format progress

bastelfreak:
For the past decade, people used a .rspec_parallel to configure this. We can set it directly in the rake task. That enables us to remove the .rspec_parallel and ultimately reduces the number of files required for pdk update/modulesync.

## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -111,8 +111,9 @@ task :parallel_spec_standalone do |_t, args|
     warn 'No files for parallel_spec to run against'
   else
 
-    args = ['-t', 'rspec']
-    args.push('--').concat(ENV['CI_SPEC_OPTIONS'].strip.split).push('--') unless ENV['CI_SPEC_OPTIONS'].nil? || ENV['CI_SPEC_OPTIONS'].strip.empty?
+    args = ['--type', 'rspec']
+    additional_options = ['--format', 'progress'] + ENV['CI_SPEC_OPTIONS'].to_s.strip.split
+    args.push('--').concat(additional_options).push('--')
     args.concat(Rake::FileList[pattern].to_a)
 
     ParallelTests::CLI.new.run(args)


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -111,8 +111,13 @@
...
   else
 
-    args = ['-t', 'rspec']
-    args.push('--').concat(ENV['CI_SPEC_OPTIONS'].strip.split).push('--') unless ENV['CI_SPEC_OPTIONS'].nil? || ENV['CI_SPEC_OPTIONS'].strip.empty?
+    args = ['--type', 'rspec']
+    additional_options = if ENV['CI_SPEC_OPTIONS'].nil? || ENV['CI_SPEC_OPTIONS'].strip.empty?
+                           ['--format', 'progress']
+                         else
+                           ENV['CI_SPEC_OPTIONS'].strip.split + ['--format', 'progress']
+                         end
```
ekohl, line 119: Isn't this equivalent?
```suggestion
    additional_options = ENV['CI_SPEC_OPTIONS'].to_s.strip.split + ['--format', 'progress']
```

Thought I'd suggest to alter the order so user provided options come last. I think rspec prefers the later options so users could still override the format:
```suggestion
    additional_options = ['--format', 'progress'] + ENV['CI_SPEC_OPTIONS'].to_s.strip.split
```

bastelfreak: yes, but `ENV['CI_SPEC_OPTIONS'].to_s.strip.split` will fail if `ENV` returns nil.

ekohl: `nil.to_s` returns an empty string (`""`)

bastelfreak: :thinking: let me test again

jordanbreen28:
@bastelfreak 👋 dropping a note to say we aren't ignoring this. I'll update the thread again shortly.

gavindidrichsen (approved):
Looks good to us!  Thanks.

PR #447, opened 2024-04-18
(CAT-273) Remove plan exclusion from rake tasks

LukasAud:
The plan exclusion was originally added to the rake tasks due to Bolt not being able to handle syntax for them. This was the case 6 years ago but not anymore. Therefore, we can remove the exclusion from the rake.

Original PR: https://github.com/puppetlabs/puppetlabs_spec_helper/pull/270

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [x] Manually verified.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -191,7 +191,6 @@ PuppetSyntax.exclude_paths << 'spec/fixtures/**/*'
 PuppetSyntax.exclude_paths << 'pkg/**/*'
 PuppetSyntax.exclude_paths << 'vendor/**/*'
 PuppetSyntax.exclude_paths << '.vendor/**/*'
-PuppetSyntax.exclude_paths << 'plans/**/*'
 PuppetSyntax.check_hiera_keys = true
 PuppetSyntax.check_hiera_data = true
 


lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -191,7 +191,6 @@
 PuppetSyntax.exclude_paths << 'pkg/**/*'
 PuppetSyntax.exclude_paths << 'vendor/**/*'
 PuppetSyntax.exclude_paths << '.vendor/**/*'
-PuppetSyntax.exclude_paths << 'plans/**/*'
```
bastelfreak, line 194: That will start syntax checks for plans, right? Does that qualify for a major release?

LukasAud: I initially placed a maintenance label as I wasnt entirely sure about the reach of this change but I had a short chat with the team and it sounds like it would warrant a Y release (if we treat it like expanded functionality).

Do you have any strong opinions on having a major release for this?

bastelfreak: If this makes pipeline fail that passed previously I think it's a major change.

LukasAud: 🤔

bastelfreak: I'm fine with what @ekohl said.

gavindidrichsen (approved):
LGTM

PR #452, opened 2024-05-09
spec_standalone: apply default formatting

bastelfreak:
usually we've a .rspec file in every module:
https://github.com/puppetlabs/pdk-templates/blob/main/moduleroot/.rspec.erb

By moving this to the rake task we can delete the whole .rspec file. Similar to https://github.com/puppetlabs/puppetlabs_spec_helper/pull/446


bastelfreak:
@gavindidrichsen can you review this one as well?

gavindidrichsen (approved):
Thanks!

david22swan:
@bastelfreak Look's like this will have to be reverted unfortunately. It's causing knock on failures within the PDK when the output is set as junit: https://github.com/puppetlabs/pdk/actions/runs/9376052167/job/25816947136#step:6:1558

It seems like we the PDK relies on this output being given in the basic form.

PR #472, opened 2024-09-11
(breaking) - Remove build rake tasks

jordanbreen28:
## Summary
this removes the build rake tasks from puppetlabs_spec_helper, as
it does not make sense to have these exist in a gem which sole purpose
is to help unit testing.

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -121,11 +121,6 @@ task :parallel_spec_standalone do |_t, args|
   end
 end
 
-desc 'Clean a built module package'
-task :clean do
-  FileUtils.rm_rf('pkg/')
-end
-
 require 'puppet-lint/tasks/puppet-lint'
 # Must clear as it will not override the existing puppet-lint rake task since we require to import for
 # the PuppetLint::RakeTask
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -121,30 +121,6 @@ task :parallel_spec_standalone do |_t, args|
   end
 end
 
-desc 'Build puppet module package'
-task :build do
-  Rake::Task['build:pdk'].invoke
-end
-
-namespace :build do
-  desc 'Build Puppet module with PDK'
-  task :pdk do
-    require 'pdk/util'
-    require 'pdk/module/build'
-
-    path = PDK::Module::Build.invoke(force: true, 'target-dir': File.join(Dir.pwd, 'pkg'))
-    puts "Module built: #{path}"
-  rescue LoadError
-    _ = `pdk --version`
-    unless $CHILD_STATUS.success?
-      warn 'Unable to build module. Please install PDK or add the `pdk` gem to your Gemfile.'
-      abort
-    end
-
-    system('pdk build --force')
-  end
-end
-
 desc 'Clean a built module package'
 task :clean do
   FileUtils.rm_rf('pkg/')


jordanbreen28:
@bastelfreak yes please, just until we get more eyes!

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -123,26 +123,11 @@
...
-    unless $CHILD_STATUS.success?
-      warn 'Unable to build module. Please install PDK or add the `pdk` gem to your Gemfile.'
-      abort
-    end
-
-    system('pdk build --force')
-  end
+  require 'puppet/modulebuilder'
+  target_dir = File.join(Dir.pwd, 'pkg')
+  builder = Puppet::Modulebuilder::Builder.new(Dir.pwd, target_dir, nil)
```
ekohl, line 128: Logger already defaults to `nil`
```suggestion
  builder = Puppet::Modulebuilder::Builder.new(Dir.pwd, target_dir)
```

And `target_dir` defaults to `File.join(source, 'pkg')` so you could even simplify it to:
```suggestion
  builder = Puppet::Modulebuilder::Builder.new(Dir.pwd)
```

ekohl (changes requested):
I'd just drop them altogether since there's already the same code in `puppet-blacksmith` to provide `module:build`:
https://github.com/voxpupuli/puppet-blacksmith/blob/3f187bafe5c5785ff6bafe9f9ca06410109e0b03/lib/puppet_blacksmith/rake_tasks.rb#L57-L63

This was done in https://github.com/voxpupuli/puppet-blacksmith/commit/f955a02cb7621021530233d8a1ba3d18b331e017 about 4 years ago.

lib/puppetlabs_spec_helper/rake_tasks.rb
jordanbreen28: Thanks @ekohl, I've updated!

jordanbreen28:
> I'd just drop them altogether since there's already the same code in `puppet-blacksmith` to provide `module:build`: https://github.com/voxpupuli/puppet-blacksmith/blob/3f187bafe5c5785ff6bafe9f9ca06410109e0b03/lib/puppet_blacksmith/rake_tasks.rb#L57-L63
> 
> This was done in [voxpupuli/puppet-blacksmith@f955a02](https://github.com/voxpupuli/puppet-blacksmith/commit/f955a02cb7621021530233d8a1ba3d18b331e017) about 4 years ago.

I think for now we'll leave the build rake task in, by default puppetlabs_spec_helper is pulled into modules by the pdk templates and therefore available to all pdk users without the need to add another dependency.

ekohl (commented):
I still lean to complete removal. I think most people rely on puppet-blacksmith and that already has the integration

bastelfreak:
I agree with @ekohl here. puppetlabs_spec_helper could also just include puppet-blacksmith, so users don't have to do it?

jordanbreen28:
@ekohl @bastelfreak yeah I realised that after I updated, the joys of working before you've had your morning coffee. I suppose no time like the present to implement a big change like this, I'm going to raise it with the team.

ekohl:
> could also just include puppet-blacksmith, so users don't have to do it?

I'm not sure I agree with this. If you look at the name it's the spec helper: designed for (unit) tests. Not the release helper.

With Vox Pupuli we put the release gems in their own group so you can install a more minimal set of gems. If `puppetlabs_spec_helper` depends on `puppet-blacksmith` that becomes impossible.

jordanbreen28:
@ekohl after some discussion with the team, we agree with your point. The `build` rake tasks do not really belong here in puppetlabs_spec_helper, and as such I'm going to update this PR to action there removal (thanks for the help with that!).

Moving forward, our next item is to add a build rake task to puppet-modulebuilder. We appreciate that `puppet-blacksmith` includes this already, but believe for our use-cases it makes sense for us to rely upon our own tooling.

lib/puppetlabs_spec_helper/rake_tasks.rb
```diff
@@ -121,30 +121,6 @@
...
-    unless $CHILD_STATUS.success?
-      warn 'Unable to build module. Please install PDK or add the `pdk` gem to your Gemfile.'
-      abort
-    end
-
-    system('pdk build --force')
-  end
-end
-
 desc 'Clean a built module package'
```
ekohl, line 124: Does this still belong here? If the build part is removed. There's already a separate `spec_clean` that removes the fixtures

ekohl (commented):
I started to extract all the fixture handling to https://github.com/ekohl/puppet_fixtures too, but never got around to completing it. My intention with that was that `puppetlabs_spec_helper` becomes a really pure combination of building blocks. For Vox Pupuli I was hoping to remove `puppetlabs_spec_helper` from our toolchain entirely.

My reasoning is that I need fixtures in our acceptance tests, but puppetlabs_spec_helper is mostly focused on unit tests. Once you extract fixtures, there's really very little left in this gem that we need.

ekohl:
> @ekohl after some discussion with the team, we agree with your point. The `build` rake tasks do not really belong here in puppetlabs_spec_helper, and as such I'm going to update this PR to action there removal (thanks for the help with that!).

Note this gem already loads in puppet-blacksmith if available:
https://github.com/puppetlabs/puppetlabs_spec_helper/blob/6e600d1bf4f11d3c87f0fa34cd75d1a4ae2cf173/lib/puppetlabs_spec_helper/rake_tasks.rb#L20-L24

You can easily end up defining the task twice. I think rake tasks are additive, so you could end up building the module twice.

Since it's a soft dependency, you can't declare you're incompatible with certain versions. Luckily puppet-blacksmith does depend on puppet-modulebuilder so you could make it an incompatible version bump there and update puppet-blackmsith accordingly.

So it's not impossible, but it's not as trivial as adding a new method.

lib/puppetlabs_spec_helper/rake_tasks.rb
jordanbreen28: No, you're right. Nice spot, thanks!

jordanbreen28:
> Note this gem already loads in puppet-blacksmith if available:
> 
> You can easily end up defining the task twice. I think rake tasks are additive, so you could end up building the module twice.
> 
> Since it's a soft dependency, you can't declare you're incompatible with certain versions. Luckily puppet-blacksmith does depend on puppet-modulebuilder so you could make it an incompatible version bump there and update puppet-blackmsith accordingly.
> 
> So it's not impossible, but it's not as trivial as adding a new method.

@ekohl those are all valid points, thanks for going through all that.
From what I have gathered, it may just be better to leave `puppet-blacksmith` here as a soft dependency, that way the tasks it offers will be available to users if the gem is present.

So my thinking here is:
1. Remove build tasks from spec_helper (as we have done)
2. Keep soft dependency of `puppet-blacksmith` in puppetlabs_spec_helper. As you have already mentioned, `puppet-blacksmith` does not make sense as a hard dependency here..
3. For PDK users, we can add puppet-blacksmith to the pdk-templates so the build tasks will still be present for users who use pdk to manage their modules.

Does this sound more inline with what you're thinking?

ekohl (approved):
> Does this sound more inline with what you're thinking?

Yes, very much.

danadoherty639 (approved):

PR #478, opened 2024-10-21
(maint) Add note to update the pdk-templates lint whenever it is changed here

david22swan:
In order to ensure that our lint testing remains in step we must keep these two files in sync with eachother:
- https://github.com/puppetlabs/puppetlabs_spec_helper/blob/main/lib/puppetlabs_spec_helper/rake_tasks.rb#L124
- https://github.com/puppetlabs/pdk-templates/blob/d9e6478ff7730a8ce072ce0c1d36a9b79f05ba2c/config_defaults.yml#L69

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [x] Manually verified.

### DIFF
--- a/lib/puppetlabs_spec_helper/rake_tasks.rb
+++ b/lib/puppetlabs_spec_helper/rake_tasks.rb
@@ -121,6 +121,8 @@ task :parallel_spec_standalone do |_t, args|
   end
 end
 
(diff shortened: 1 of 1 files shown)


gavindidrichsen (approved):
Thanks @david22swan !

PR #481, opened 2025-02-25
(CAT-2113) Adjusting minimum ruby version required

LukasAud:
Following investigation on an issue that prevented puppet-resource_api from consuming latest puppetlabs_spec_helper, it was noted that this repository assumes Ruby 2.7 as the minimum required version, causing some repositories to fail their testing environment builds due to the fact that they are consuming some form of Ruby 2.6 in order to test properly against puppetserver.

This commit adjusts the required Ruby version to the standard used across other modules/tools of the ecosystem.


bastelfreak:
Why does the resource API still need to support Ruby 2.6 but also latest puppetlabs_spec_helper?

.rubocop_todo.yml
```diff
@@ -1,11 +1,18 @@
...
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
+# Offense count: 1
+# Configuration parameters: Severity, Include.
+# Include: **/*.gemspec
+Gemspec/RequiredRubyVersion:
+  Exclude:
+    - 'puppetlabs_spec_helper.gemspec'
```
ekohl, line 14: I'd consider this poor practice. You are telling RuboCop to assume you can use Ruby 2.7 features while being Ruby 2.6 compatible. It's just a matter of time before that breaks IMHO.

ekohl (changes requested):

.rubocop_todo.yml
bastelfreak: Yes. I think the resource-api needs to drop Ruby 2.6 support

ekohl: I think you misunderstood the issue. IIRC the real issue is that JRuby used in Puppetserver 7 is officially only Ruby 2.6 compatible so I'm guessing some of Perforce's internal CI jobs run with Ruby 2.6.

bastelfreak: Yes. But that's EoL in three days: https://www.puppet.com/products/puppet-enterprise/support-lifecycle. So why does resource-api *now* need to use latest puppetlabs_spec_helper + Ruby 2.6?

ekohl: That's something I can't answer.

LukasAud:
@bastelfreak @ekohl So my idea was to release one last minor version compatible with 2.6 before major changes happen once Puppet 7 is EOL (a "just in case someone needs it" scenario).

After chatting with my team, we agreed it makes sense to just wait until P7 EOL and just remove the unsupported Ruby testing. So I'll be closing down this PR.

PR #485, opened 2026-07-22
(MODULES-11700) Allow puppet-lint 5.x (Ruby 3.4+ compatible)

imaqsood:
## What

Pin the `puppet-lint` runtime dependency to the 5.x line so it works on Ruby 3.4+ / Puppet 9:

```diff
-  spec.add_runtime_dependency 'puppet-lint', '~> 4.0'
+  spec.add_runtime_dependency 'puppet-lint', '~> 5.0'
```

## Why

puppet-lint 4.x crashes on Ruby 3.4+; 5.x is the Ruby 3.4+ / Puppet 9 compatible line and is also compatible with the Puppet 8 pipelines. Per review, puppet-lint 4 and 5 differ internally (different lint results), so this requires one line rather than allowing both.

## Context

Unblocks the Puppet 9 support work in `puppetlabs-apache` (MODULES-11700), whose Gemfile currently pins this branch on the Puppet 9 lint lane.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/puppetlabs_spec_helper.gemspec
+++ b/puppetlabs_spec_helper.gemspec
@@ -30,7 +30,7 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency 'mocha', '>= 1.0', '< 3'
   spec.add_runtime_dependency 'pathspec', '>= 0.2', '< 3'
-  spec.add_runtime_dependency 'puppet-lint', '>= 4.0', '< 6.0'
+  spec.add_runtime_dependency 'puppet-lint', '~> 5.0'
   spec.add_runtime_dependency 'puppet-syntax', '~> 4.1', '>= 4.1.1'
   spec.add_runtime_dependency 'rspec-github', '>= 2.0', '< 4'
   spec.add_runtime_dependency 'rspec-puppet', '~> 5.0'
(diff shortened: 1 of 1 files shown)


puppetlabs_spec_helper.gemspec
```diff
@@ -30,7 +30,7 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency 'mocha', '>= 1.0', '< 3'
   spec.add_runtime_dependency 'pathspec', '>= 0.2', '< 3'
-  spec.add_runtime_dependency 'puppet-lint', '~> 4.0'
+  spec.add_runtime_dependency 'puppet-lint', '>= 4.0', '< 6.0'
```
bastelfreak: I don't recommend this. Some internals changed in puppet-lint 4 vs 5. You will get different linting results. I highly recommend to either pin to puppet-lint 5, or even better, pull in voxpupuli-puppet-lint-plugins, which also brings you suitable linter plugins.

bastelfreak (changes requested):

puppetlabs_spec_helper.gemspec
david22swan: I would agree, it needs to be one version or the other

david22swan: As it stands, puppet-lint v5 should be compatible with the Puppet 8 pipelines

imaqsood:
Thanks @bastelfreak @david22swan — good point on the 4-vs-5 internals. Changed to pin `puppet-lint` to `~> 5.0` (one line, not both) in c2808e1. CI resolves to 5.x already so it stays green, and per the note puppet-lint 5 is compatible with the Puppet 8 pipelines too.

david22swan (approved):
LGTM, may need to wait a bit on release as we plan out what all thats needs to go into the next one to get it fully Ruby 4 compatible but it should be out soon enough.

